PR #9, opened 2026-01-21
Add early dirty worktree check before branch operations

melonamin:
## Summary

- Add `IsDirty()` method to `pkg/git` that detects staged changes or modified tracked files
- Call `IsDirty()` at the start of `createBranchIfNeeded()` to fail early with a clear error message
- When worktree is dirty, user sees: `worktree has uncommitted changes, commit or stash before running ralphex`

Previously, go-git's `wt.Checkout()` would fail with a cryptic error when attempting to switch branches with uncommitted changes.

## Test Plan

- [x] `make test` passes (78.5% coverage)
- [x] `make lint` passes (no new warnings)
- [x] Manual test: dirty worktree shows clear error message
- [x] Manual test: staged file shows clear error message  
- [x] Manual test: clean worktree proceeds with branch creation

### DIFF
--- a/pkg/git/git_test.go
+++ b/pkg/git/git_test.go
@@ -474,6 +474,21 @@ func TestRepo_IsDirty(t *testing.T) {
 		assert.True(t, dirty)
 	})
 
+	t.Run("deleted tracked file returns true", func(t *testing.T) {
+		dir := setupTestRepo(t)
+		repo, err := Open(dir)
+		require.NoError(t, err)
+
+		// delete the existing README.md (which is tracked)
+		readmePath := filepath.Join(dir, "README.md")
+		err = os.Remove(readmePath)
+		require.NoError(t, err)
+
+		dirty, err := repo.IsDirty()
+		require.NoError(t, err)
+		assert.True(t, dirty)
+	})
+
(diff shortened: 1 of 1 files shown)


umputun:
LGTM, clean implementation. one minor suggestion for test coverage:

consider adding a test for deleted tracked file (unstaged):

```go
t.Run("deleted tracked file returns true", func(t *testing.T) {
    dir := setupTestRepo(t)
    repo, err := Open(dir)
    require.NoError(t, err)

    // delete the existing README.md (which is tracked)
    readmePath := filepath.Join(dir, "README.md")
    err = os.Remove(readmePath)
    require.NoError(t, err)

    dirty, err := repo.IsDirty()
    require.NoError(t, err)
    assert.True(t, dirty)
})
```

the code already handles this via `s.Worktree == git.Deleted`, just no explicit test for it. not a blocker.

linnik:
`func (r *Repo) CheckoutBranch(name string) error` shouldn't silently delete untracked files in the first place?
@umputun what do you think about changing its behavior to `git.CheckoutOptions{Branch: branchRef, Keep: true}`

PR #10, opened 2026-01-21
fix: support git worktrees

melonamin:
## Summary
- Use `PlainOpenWithOptions` with `EnableDotGitCommonDir: true` to properly handle git worktrees
- Fixes error `reference not found` when running ralphex from a git worktree

## Problem
When running ralphex from a git worktree, it fails with:
```
error: get current branch: get HEAD: reference not found
```

This happens because git worktrees use a different `.git` structure - instead of a directory, `.git` is a file pointing to the actual git directory in the main repository.

## Solution
Enable `EnableDotGitCommonDir` option in go-git's `PlainOpenWithOptions`, which tells go-git to follow the `commondir` indirection and find shared refs (like HEAD) in the main repository.

## Test plan
- [x] Unit tests pass (`make test`)
- [x] Linter passes on git package (`golangci-lint run pkg/git/...`)
- [x] Verified fix works in actual git worktree (see below)

### Verification
Created test that demonstrates the fix:
```
SUCCESS: Branch = feature-branch
Without fix: Head() failed: reference not found (THIS IS THE BUG)
```

Also verified regular repos still work:
```
SUCCESS: Branch = master
Without fix: also works (unexpected)
```

### DIFF
--- a/pkg/git/git_test.go
+++ b/pkg/git/git_test.go
@@ -2,6 +2,7 @@ package git
 
 import (
 	"os"
+	"os/exec"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


umputun:
LGTM. Minimal fix with proper test coverage - creates real worktree and verifies branch detection works.

umputun:
> @umputun FYI go-git v6 [has proper API](https://pkg.go.dev/github.com/go-git/go-git/v6/x/plumbing/worktree#New) for worktrees. Not important enough to upgrade, but if there no blockers, why not?

I had no clue they have v6. As long as it doesn't break things, that's fine with me.

PR #42, opened 2026-01-29
fix: handle context cancellation during interactive input prompts

chloyka:
## Description                                                                                                                                                                                                             
                                                                                                                                                                                                                          
### Problem                                                                                                                                                                                                        
Ctrl+C and kill signals don't work when the app waits for user input (yes/no prompts, plan description, option selection). The app hangs and ignores SIGINT/SIGTERM.                                                    
This is counterintuitive — you can't kill the app even with kill because it's blocked on stdin read   

### Solution
Pass context to all input functions and use ReadLineWithContext() that respects context cancellation.

### DIFF
--- a/pkg/input/input.go
+++ b/pkg/input/input.go
@@ -29,6 +29,10 @@ func ReadLineWithContext(ctx context.Context, reader *bufio.Reader) (string, err
 		return "", fmt.Errorf("read line: %w", err)
 	}
 
+	if err := ctx.Err(); err != nil {
+		return "", fmt.Errorf("read line: %w", err)
+	}
+
 	go func() {
 		line, err := reader.ReadString('\n')
 		resultCh <- readLineResult{line: line, err: err}


umputun:
couple linter issues:

1. **misspelling** - `cancelled` → `canceled` (US spelling) in 6 places:
   - `cmd/ralphex/main_test.go:69,70,1426`
   - `pkg/input/input.go:23,155`
   - `pkg/input/input_test.go:140`

2. **wrapcheck** - `ctx.Err()` should be wrapped at `pkg/input/input.go:35`:
   ```go
   // instead of
   return "", ctx.Err()
   // use
   return "", fmt.Errorf("read line: %w", ctx.Err())
   ```

**note on goroutine lifecycle** (not a blocker, just FYI): when context is cancelled, the goroutine in `ReadLineWithContext` stays blocked on `reader.ReadString()` until the read completes or process exits. this is the standard pattern for cancellable blocking I/O in Go - there's no way to actually interrupt a syscall. fine for this use case since the app is exiting anyway.

umputun (approved):
LGTM, thx for addressing all the feedback

PR #75, opened 2026-02-08
refactor: make PhaseHolder single source of truth for execution phase

ZhilinS:
Sometimes it's too pricey to always use Claude Opus models for planning, execution and review, especially on the $100 plan, where session limits tend to end in an hour. This PR allows configuring the model for each phase separately. 

- Added functionality to set models in config for each phase separately with config options (`claude_model_task`, `claude_model_review`, `claude_model_plan`)
- When set, ralphex passes `--model <value>` to the claude CLI for that phase; when unset, claude uses its default
- Refactored phase holding logic, consolidated in a single `PhaseHolder`, replacing scattered phase tracking in logger and broadcast logger
- Added callback to phase change for backward compatibility with the BroadcastLogger

### DIFF
--- a/pkg/config/values.go
+++ b/pkg/config/values.go
@@ -36,6 +36,7 @@ type Values struct {
 	FinalizeEnabledSet   bool // tracks if finalize_enabled was explicitly set
 	PlansDir             string
 	WatchDirs            []string // directories to watch for progress files
+
 	// notification settings
 	NotifyChannels        []string // channels to use: telegram, email, webhook, slack, custom
 	NotifyChannelsSet     bool     // tracks if notify_channels was explicitly set (allows empty to disable)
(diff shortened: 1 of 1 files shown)


ZhilinS:
Thanks for the in-depth review.

I've stripped out the per-phase model selection code; the PR now contains only the PhaseHolder refactoring.

On the model control approach - I'm usually not that fine-grained about model selection for different claude tasks myself, but I agree that the agent frontmatter idea is better. Since agent frontmatter isn't supported yet, I'll implement that in a follow-up PR.
  
One concern about the plan phase: sometimes the task is subtle enough that firing up opus isn't necessary. A `--model` CLI flag for `--plan` mode would solve this, but that's for the next PR discussion.

PR #110, opened 2026-02-15
Fix: use Setsid instead of Setpgid to prevent terminal signal hang

paskal:
## Summary

Claude CLI processes spawned by ralphex get stopped (state T) and hang indefinitely on macOS when running from a terminal. The hang is 100% reproducible since v0.4.0 and affects all tested Claude CLI versions (v2.1.30–v2.1.42).

## Root cause

`Setpgid: true` in `setupProcessGroup()` creates a new process group within the **same session**, so the child inherits the controlling terminal. The child's process group becomes a "background process group" from the terminal's perspective, which means **any** subprocess that Claude spawns—shell commands via the Bash tool, language servers (gopls), environment snapshots (zsh)—will receive SIGTTIN/SIGTTOU signals from the kernel when it attempts terminal I/O. This stops the entire process group, hanging ralphex indefinitely.

The issue affects anything Claude's subprocess tree does that touches the terminal; it's not limited to specific child processes.

## Process tree during the hang

Every observed hang looked like this—Claude and all its children in T (stopped) state, same PGID:

```
  PID  PPID  PGID STAT COMMAND
78401     1 78401 S+   ralphex --max-iterations 50 docs/plans/...
78425 78401 78425 T    claude --dangerously-skip-permissions --output-format stream-json ...
78429 78425 78425 T    gopls
78534 78425 78425 T    /bin/zsh -c -l SNAPSHOT_FILE=... source ~/.zshrc < /dev/null ...
```

File descriptors of the stuck Claude process:
```
fd 0: /dev/null (read)
fd 1: PIPE -> ralphex (stdout)
fd 2: PIPE -> ralphex (stderr merged)
fd 9,11-14: TCP ESTABLISHED -> api.anthropic.com:443
```

The zsh shell snapshot process is what triggers SIGTTIN—it runs as a login shell (`-l`) and while `< /dev/null` redirects stdin, the shell initialisation may still probe the terminal before the redirect takes effect. Once any process in the group gets stopped, the entire PGID is affected.

## Fix

Switch from `Setpgid: true` to `Setsid: true`. This creates a new **session**, fully detaching the child from the controlling terminal. No descendant can trigger SIGTTIN/SIGTTOU since there is no controlling terminal in the new session. Process group cleanup via `syscall.Kill(-pid, sig)` still works identically—the child becomes session leader of its own process group, so the negative PID kill targets the same set of processes.

## Cancellation still works

Ctrl+C graceful shutdown works the same before and after the fix:

```
# before (Setpgid)
--- task iteration 1 ---
[26-02-15 09:54:40] I'll start by reading the plan file to understand the work and find the first uncompleted task
[26-02-15 09:54:40] section.
[26-02-15 09:54:43] Now let me also check the progress file for any previous context.

interrupting... (force exit in 5s)
error: runner: task phase: claude execution: context canceled

# after (Setsid)
--- task iteration 2 ---
[26-02-15 11:32:07] I'll start by reading the plan file and the progress file to understand the current state.
[26-02-15 11:32:12] The first uncompleted task section is Task 2: Fix Stripe handlers returning plain text errors. Let me
[26-02-15 11:32:12] read the relevant source files.
[26-02-15 11:32:16] Now I need to understand the `SendErrorJSON` function signature. Let me check it.

interrupting... (force exit in 5s)
error: runner: task phase: claude execution: context canceled
```

## Investigation

Binary search across ralphex versions confirmed the breaking change:

| Version | Setpgid | Result |
|---------|---------|--------|
| v0.3.0  | no      | works  |
| v0.4.0  | yes     | hangs  |
| v0.8.0+ | yes     | hangs  |
| v0.12.0 | yes     | hangs  |

Symptoms observed across 6 consecutive hangs:
- Claude process enters T (stopped) state within minutes
- All children (gopls, zsh) also in T state, same PGID
- JSONL session log ends with `stop_reason: null`—API stream never completes
- TCP connections to Anthropic API remain ESTABLISHED but idle

Shell-level reproduction scripts (including full setsid+pipe+devnull) did **not** reproduce the hang, confirming the issue is specific to Go's `Setpgid` within `SysProcAttr` keeping the child in the parent's session.

### DIFF
--- a/pkg/executor/procgroup_unix.go
+++ b/pkg/executor/procgroup_unix.go
@@ -24,11 +24,14 @@ type processGroupCleanup struct {
 	err  error
 }
 
-// setupProcessGroup configures command to run in its own process group.
-// This allows killing all descendant processes when cleanup is needed.
+// setupProcessGroup configures command to run in its own session and process group.
+// uses Setsid to create a new session, fully detaching the child from the parent's
+// controlling terminal. this prevents SIGTTIN/SIGTTOU signals that stop the child
+// process group when its descendants (e.g. shell, gopls) attempt terminal I/O.
+// the child becomes session leader of its own process group, so killing via -pid still works.
 // Must be called before cmd.Start().
 func setupProcessGroup(cmd *exec.Cmd) {
-	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setsid: true}
 }
 
 // newProcessGroupCleanup creates a cleanup handler for the given command.


umputun (approved):
lgtm, solid investigation and clean fix.

one minor thing - `procgroup_windows.go:22` still references `Setpgid` in the comment, should mention `Setsid` now (or just say "unix session/process group APIs"). not a blocker.

re Copilot's suggestion to capitalize the comment sentences - ignore it, lowercase is the project convention for non-godoc continuation lines.

PR #111, opened 2026-02-15
feat: add SELinux support for Docker volume mounts

jjdoughnut:
### Summary
Auto-detects **SELinux** via /**sys/fs/selinux/enforce** and appends the **:z** shared label to all Docker bind mounts in **ralphex-dk.sh.** Read-only mounts get **:ro,z.** No-op on non-SELinux systems. Includes mocked tests that exercise both code paths on any platform. 

### Comment
I couldn't start ralphex using ralphex-dk.sh on Fedora 43. I was getting this error:
```
error: open git repo: new git service: open git repository /workspace: fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```
I used ralphex to debug and build the fix. Not sure I like how it split tests to **with** and **without_selinux** prefixes, but I also don't know how to do that better.

### DIFF
--- a/scripts/ralphex-dk.sh
+++ b/scripts/ralphex-dk.sh
@@ -21,6 +21,7 @@ import sys
 import tempfile
 import threading
 import unittest
+import unittest.mock
 from pathlib import Path
 from types import FrameType
 from typing import Optional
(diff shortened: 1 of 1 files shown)


umputun (approved):
lgtm, thx. clean fix for a real problem on SELinux-enabled distros.

`:z` is the right choice (shared label), detection via `/sys/fs/selinux/enforce` is standard, no-op on non-SELinux systems. tests look good.

PR #152, opened 2026-02-24
fix: make web dashboard reachable from host in Docker

### LINKED ISSUE #151: Web dashboard unreachable from host when running in Docker
The web dashboard binds to `127.0.0.1` (`pkg/web/server.go:89`), making it unreachable from the host when running via the Docker wrapper despite `-p 8080:8080` port mapping. Docker forwards traffic to the container's bridge interface, not loopback.

**Suggestion:** Add a `--host` flag (default `127.0.0.1`) and have the Docker wrapper pass `--host 0.0.0.0` when `--serve` is used. This keeps non-Docker users on loopback by default while making the dashboard work in Docker.

nnemirovsky:
## Summary
- Add `--host` flag (default `127.0.0.1`, env `RALPHEX_WEB_HOST`) to configure the dashboard bind address
- Docker wrapper sets `RALPHEX_WEB_HOST=0.0.0.0` when `--serve` is used
- Non-Docker users stay on loopback by default

Fixes #151

## Problem
The dashboard binds to `127.0.0.1` inside the container, which is unreachable from the host despite `-p 8080:8080` port mapping. Docker forwards traffic to the container's bridge interface, not its loopback.

### DIFF
--- a/pkg/web/dashboard.go
+++ b/pkg/web/dashboard.go
@@ -16,9 +16,9 @@ import (
 const serverStartupTimeout = 100 * time.Millisecond
 
 // ConnectHost returns the host to use in user-facing URLs.
-// maps wildcard addresses (0.0.0.0, ::) to localhost since those aren't connectable.
+// maps wildcard and loopback addresses to localhost since those aren't connectable.
 func ConnectHost(host string) string {
-	if host == "" || host == "0.0.0.0" || host == "::" {
+	if host == "" || host == "0.0.0.0" || host == "::" || host == "::1" || host == "127.0.0.1" {
 		return "localhost"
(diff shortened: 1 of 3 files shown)


umputun (approved):
lgtm, clean and focused fix. good that the Docker host-side port is now bound to `127.0.0.1:{port}:8080` instead of all interfaces, and the env var check in the wrapper allows user override.

one minor nit - `http://%s:%d` in the log URLs would produce unparseable output for raw IPv6 literals (e.g. `http://fe80::1:8080`), but since `ConnectHost` maps all realistic addresses to `localhost`, this is purely theoretical. not a blocker.

PR #154, opened 2026-02-24
fix: forward host timezone to Docker container

### LINKED ISSUE #153: Docker container uses hardcoded America/Chicago timezone instead of host timezone
The base image (`ghcr.io/umputun/baseimage/app:latest`) has `/etc/timezone` set to `America/Chicago`. The Docker wrapper doesn't forward the host's timezone, so all progress log timestamps show Chicago time regardless of the user's locale.

**Suggestion:** Have the wrapper detect the host's timezone and pass it via `-e TZ=...` to the container.

nnemirovsky:
## Summary
- The base Docker image defaults to `America/Chicago`, causing all timestamps to display in that timezone regardless of the user's locale
- The wrapper script now detects the host timezone from `$TZ`, `/etc/localtime` symlink, or `/etc/timezone` and forwards it to the container via `-e TZ=`
- Includes 4 tests for timezone detection

Closes #153

### DIFF
--- a/scripts/ralphex-dk.sh
+++ b/scripts/ralphex-dk.sh
@@ -378,16 +378,16 @@ def _detect_host_timezone() -> str:
     then falls back to reading /etc/timezone (Debian/Ubuntu).
     returns empty string if detection fails.
     """
-    # try /etc/localtime symlink (macOS and Linux)
-    try:
-        link = os.path.realpath("/etc/localtime")
-        # extract IANA name from path like /usr/share/zoneinfo/America/New_York
-        marker = "/zoneinfo/"
-        idx = link.find(marker)
-        if idx != -1:
-            return link[idx + len(marker):]
-    except OSError:
-        pass
+    # readlink first (macOS), then realpath (Linux)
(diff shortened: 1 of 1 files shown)


umputun (changes requested):
one issue - timezone detection doesn't work on macOS. `os.path.realpath("/etc/localtime")` resolves to `/usr/share/zoneinfo.default/America/Chicago` (note `zoneinfo.default`, not `zoneinfo`), so the `/zoneinfo/` marker won't match and the function silently returns `""`.

fix: try `os.readlink("/etc/localtime")` first, which returns the direct symlink target (`/var/db/timezone/zoneinfo/America/Chicago` on macOS where the marker works), then fall back to `os.path.realpath` for Linux distros where `/etc/localtime` is a regular file.

the `test_tz_from_localtime_macos_path` test should also be updated - it mocks `realpath` to return `/var/db/timezone/zoneinfo/...` but that's not what `realpath` actually returns on macOS.

everything else looks good - clean code, proper error handling, correct `$TZ` precedence.

PR #162, opened 2026-02-26
add configurable VCS command with Mercurial support

paskal:
## Summary

- add `vcs_command` config option so the git backend uses a configurable binary instead of hardcoded `git`
- include reference `scripts/hg2git.sh` translation script mapping ~15 git subcommands to Mercurial equivalents
- script uses phase-based commit logic: `hg commit` on public (creates new draft), `hg amend` on draft (folds into existing)
- add `docs/hg-support.md` with full setup guide, prompt replacements, known limitations

## Changes

**Config** (`pkg/config/`): new `VcsCommand` string field with INI parsing, merge logic, tilde expansion, tests

**Git backend** (`pkg/git/`): `externalBackend` gets `command` field replacing all 9 hardcoded `"git"` exec calls. `NewService` accepts optional `vcsCmd ...string` parameter, defaults to `"git"`

**Wiring** (`cmd/ralphex/main.go`): pass `cfg.VcsCommand` through to `openGitService` and worktree service. skip `.git` directory check when `vcs_command != "git"` - rely on `rev-parse --show-toplevel` instead

**Script** (`scripts/hg2git.sh`): bash translation covering rev-parse, symbolic-ref, status (format conversion), checkout, add, commit (phase-based), diff (numstat), check-ignore (two-stage: hg status -i for existing files, .hgignore pattern matching for non-existent), worktree (unsupported), log

**Docs**: `docs/hg-support.md` setup guide, README/CLAUDE.md/llms.txt updated with `vcs_command` option

default `vcs_command = git` produces identical behaviour - fully backward compatible


umputun (changes requested):
the Go code is clean - mechanical replacement of hardcoded `"git"` with `e.command`, variadic param on `NewService` preserves backward compat. tests pass, linter clean, no races.

a few things:

1. error messages in `pkg/git/external.go` still hardcode `"git"` - e.g. `fmt.Errorf("git %s: %s", args[0], msg)`. when using hg2git.sh this will say "git commit: ..." which is misleading. consider passing `e.command` into the format string so it shows the actual VCS command being used

2. `.git/` in `.gitignore` - looks like your dev setup artifact (git nested in hg). doesn't belong in the project gitignore

3. `scripts/hg2git.sh` check-ignore fallback - the `.hgignore` pattern matching uses `grep -E` for all patterns but doesn't track the active syntax mode. if someone has `syntax: glob` in `.hgignore`, glob patterns like `*.pyc` will be misinterpreted as regex. docs/hg-support.md mentions grep usage but should explicitly warn about glob mode failure

4. minor: `hg log -r "bookmark('$name')"` at line 81 - single quote in `$name` would break the revset expression. low risk since names come from Go code, but worth a comment

btw, most of Copilot's 11 comments are noise. the main one about empty `vcs_command` skipping the `.git` check is wrong - the condition correctly performs the check for empty/default. the glob syntax (#3 above) and injection concern (#4) are the valid ones.

umputun (approved):
lgtm, thx

PR #169, opened 2026-03-01
feat: add AWS Bedrock support for Docker wrapper

bronislav:
## Summary

Adds AWS Bedrock provider support to the Docker wrapper. Enable via `--claude-provider bedrock` flag or `RALPHEX_CLAUDE_PROVIDER=bedrock` env var. When enabled, the wrapper automatically sets `CLAUDE_CODE_USE_BEDROCK=1`, exports credentials from AWS profile using `aws configure export-credentials`, and skips macOS keychain extraction. Credentials are passed using Docker's inherit form (`-e VAR`) so values never appear in `ps` output.

All 147 unit tests pass. See `docs/bedrock-setup.md` for setup instructions.

### DIFF
--- a/docs/bedrock-setup.md
+++ b/docs/bedrock-setup.md
@@ -18,6 +18,11 @@ AWS Bedrock provides access to Claude models through your AWS account. This allo
 3. Use short-lived credentials when possible (SSO, STS)
 4. ralphex never mounts `~/.aws` - it exports only the specific credentials needed
 
+**Avoid passing secrets on the command line.** When using `-E` to pass environment variables:
+- Prefer `-E VAR` (inherit form) over `-E VAR=value` for secrets
+- Values in `-E VAR=value` are visible in `ps` output to other users on the system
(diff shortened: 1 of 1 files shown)


umputun (changes requested):
couple things to fix:

1. `_save_and_set` helper is copy-pasted into 6 test classes identically, same for `setUp`/`tearDown`. should be a shared base class

2. `TestCredentialMerge` duplicates the merge logic from `main()` instead of testing through it. if `main()` changes, these tests won't catch regressions. extract the merge into a named function and test that

3. `RALPHEX_EXTRA_ENV` docs should mention that vars passed this way are visible in `ps` output (unlike bedrock credentials via env-file). users might pass secrets through it without realizing

umputun:
also, `RALPHEX_EXTRA_ENV` is not bedrock-related - it's a generic mechanism for passing any env vars to the container. it has its own function, tests, docs, and even a separate task in the plan. should be a separate PR to keep this one focused on bedrock only

PR #179, opened 2026-03-03
feat: add RALPHEX_EXTRA_ENV support for passing environment variables to Docker

bronislav:
## Summary

- Add `RALPHEX_EXTRA_ENV` environment variable for passing env vars to container (comma-separated)
- Add `-E`/`--env` CLI flags for passing individual env vars (uppercase E to avoid conflict with ralphex's `-e`/`--external-only`)
- Support both explicit values (`VAR=value`) and name-only form (`VAR`) that inherits from host
- Emit security warnings when sensitive names (KEY, SECRET, TOKEN, etc.) have explicit values
- Validate env var names (must match `^[A-Za-z_][A-Za-z0-9_]*`)
- Document comma limitation: values containing commas require `-E` flag (env var splits on commas)

## Test plan

- [x] `make test` - passes
- [x] `make lint` - 0 issues
- [x] Unit tests for `is_sensitive_name()`, `extract_extra_env()`, `build_env_vars()`, `validate_env_entry()`
- [x] Run `python3 scripts/ralphex-dk.sh --test` - all 72 tests pass

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,18 +318,19 @@ RALPHEX_EXTRA_VOLUMES="/data:/mnt/data:ro,/models:/mnt/models" ralphex docs/plan
 
 **Extra environment variables:**
 ```bash
-# via CLI flags (can use multiple -e)
-ralphex -e DEBUG=1 -e API_KEY docs/plans/feature.md
+# via CLI flags (can use multiple -E)
+ralphex -E DEBUG=1 -E API_KEY docs/plans/feature.md
 
 # via environment variable (comma-separated)
 RALPHEX_EXTRA_ENV="DEBUG=1,LOG_LEVEL=verbose" ralphex docs/plans/feature.md
 
 # name-only form inherits value from host (recommended for secrets)
 export API_KEY=secret123
-ralphex -e API_KEY docs/plans/feature.md
-```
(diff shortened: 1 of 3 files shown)


umputun (changes requested):
couple things:

1. **`-e` flag shadows ralphex's `--external-only`** — please change to `-E` (uppercase) to avoid breaking existing Docker wrapper users. `-e` is already taken by ralphex itself

2. **comma-in-value silently corrupts** — `RALPHEX_EXTRA_ENV="FOO=hello,world"` splits into `FOO=hello` + `world` (treated as name-only var). document this limitation somewhere (README or docstring), noting that values with commas need the `-E` flag form

3. **inconsistent validation** — `RALPHEX_EXTRA_ENV` entries go through `ENV_VAR_PATTERN` validation but CLI `-E` values don't. `-E "=bad"` would pass through unvalidated to Docker. should validate both paths the same way

4. **duplicated warning message** — sensitive name warning string appears at both line 377 and 609. extract to a constant or shared helper

otherwise the implementation is clean — follows existing patterns well, thorough tests (20 new), good edge case handling in `is_sensitive_name`

PR #199, opened 2026-03-11
feat: add opencode-as-claude wrapper and opencode-review scripts#5

mschedrin:
Add scripts to use OpenCode as an alternative provider for ralphex:
- opencode-as-claude.sh: wrapper that translates Claude Code stream-json protocol to OpenCode CLI, with auto-allow permissions and signal echo prevention
- opencode-review.sh: custom review script for external review phase
- Tests for both scripts with comprehensive coverage


umputun (approved):
all 6 items addressed, thx. I'll update README with opencode mention separately after merge.

minor note: `opencode-review.sh` still uses `echo "$var" | jq` in a few places (lines 39, 42, 48, 53) — not a blocker since those are constructed values, but worth aligning with the main wrapper's `printf '%s\n'` style for consistency.

PR #208, opened 2026-03-13
feat: add --exec flag to the docker wrapper script

bronislav:
Add `--exec` flag to open exec arbitrary command inside container instead of ralphex. 

It is mainly useful to troubleshoot custom ralphex container preserving all extra env variable and volumes that already configured for given project

### DIFF
--- a/scripts/ralphex-dk.sh
+++ b/scripts/ralphex-dk.sh
@@ -812,7 +812,11 @@ def run_docker(image: str, port: str, volumes: list[str], env_vars: list[str], b
     cmd.extend(volumes)
     cmd.extend(["-w", "/workspace"])
     if exec_cmd:
-        cmd.extend([image] + shlex.split(exec_cmd))
+        try:
+            cmd.extend([image] + shlex.split(exec_cmd))
+        except ValueError as e:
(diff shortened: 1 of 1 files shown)


bronislav:
I like `--dry-run` idea. I'll update it tomorrow.

bronislav:
Good point about the simplicity, and you're right that `--dry-run` would be smaller. I considered it but ran into a usability issue:

The wrapper does credential extraction (keychain → env, AWS profile export → env) before running docker. Credentials are passed by name inheritance (-e AWS_ACCESS_KEY_ID not -e AWS_ACCESS_KEY_ID=xxx), so they won't appear in the printed command - which is good for security.

But that means the printed command isn't directly runnable. User copies it, pastes in their shell, and gets auth failures because their shell doesn't have the extracted credentials. The command looks valid but silently fails - potentially more confusing than helpful.

Options I see:
  1. Print explicit credential values - works but security concern (history, screen sharing, logs)
  2. Print with comments explaining which vars need manual setup - adds complexity, still not copy-paste-able
  3. Keep --exec - handles credentials automatically, user gets working shell

Happy to implement `--dry-run` if you think the tradeoff is acceptable (maybe with a warning that env vars need manual setup?), or if there's another approach I'm missing.

PR #220, opened 2026-03-16
 fix: pass Claude prompt via stdin to avoid Windows cmd.exe length limit

stanurkov:
Summary

Windows compatibility: pass Claude prompt via stdin instead of -p argument.

Problem

Claude Code is a Node.js application that internally uses child_process.exec on
Windows, which routes through cmd.exe. The cmd.exe command-line limit is 8191
characters. The make_plan.txt prompt after variable expansion reaches ~8100 chars;
together with other flags (~115 chars) this exceeds the limit, causing claude to
exit immediately with "The command line is too long." and status 1.

Fix

execClaudeRunner gains a stdin io.Reader field. When ClaudeExecutor.Run() creates
the real runner (cmdRunner field is nil), it now passes the prompt via cmd.Stdin
and adds --print instead of -p <prompt>. Claude reads from stdin when stdin is a
pipe. This keeps the command line under 200 characters regardless of prompt size.

Mock runners used in tests continue to receive -p as an argument, so all existing
test assertions remain valid.

A logCmdLineStats() helper is also added: it logs the estimated command-line length
and per-argument breakdown to stderr before every process launch, making this class
of error immediately diagnosable in the future.

Companion fixes included in this PR:

- progress.go: use os.Truncate (path-based) instead of f.Truncate (fd-based) when
  truncating a completed progress file on restart. fd-based truncation fails with
  "access is denied" on Windows when the fd is opened with O_APPEND (which does not
  grant FILE_WRITE_DATA). The lock is held at this point so path/fd identity is
  stable, making path-based truncation safe here.

- progress_test.go: use filepath.ToSlash() in path assertions to handle Windows
  backslash separators.

- gemini-as-claude_test.sh, opencode-as-claude_test.sh: fix pipeline order in the
  stdin test (echo | bash, not bash | echo).

- codex-as-claude_test.sh: add stdin regression test matching the gemini/opencode
  test coverage.

Test plan

make test passes
make lint passes
GOOS=windows GOARCH=amd64 go build ./... succeeds
Manual: run ralphex --plan "..." on Windows - no longer hits "The command line is too long."

### DIFF
--- a/pkg/executor/executor_test.go
+++ b/pkg/executor/executor_test.go
@@ -769,7 +769,7 @@ func TestHelperProcess(t *testing.T) {
 	}
 	data, _ := io.ReadAll(os.Stdin)
 	fmt.Print(string(data))
-	os.Exit(0) //nolint:revive // intentional: subprocess helper must exit immediately
+	os.Exit(0)
 }
 
 // TestHelperProcessStreamJSON is not a real test — used as a subprocess by
@@ -783,7 +783,7 @@ func TestHelperProcessStreamJSON(t *testing.T) {
 	fmt.Printf(`{"type":"content_block_delta","delta":{"type":"text_delta","text":%q}}`, string(data))
 	fmt.Println()
 	fmt.Println(`{"type":"result","result":""}`)
-	os.Exit(0) //nolint:revive // intentional: subprocess helper must exit immediately
+	os.Exit(0)
 }
 
 func TestClaudeExecutor_Run_RealRunner_StdinWired(t *testing.T) {
--- a/pkg/progress/progress.go
+++ b/pkg/progress/progress.go
@@ -187,11 +187,11 @@ func NewLogger(cfg Config, colors *Colors, holder *status.PhaseHolder) (*Logger,
 
 	// if the file has a completion footer from a previous run, truncate and start fresh.
 	// this prevents mixing unrelated content when the same plan filename is reused.
-	// reads from the locked fd directly to avoid TOCTOU path-vs-inode mismatch.
+	// the file lock is held here, so path and fd refer to the same inode; path-based
+	// truncation is safe. os.Truncate (path-based) is used instead of f.Truncate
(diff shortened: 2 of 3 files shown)


umputun (changes requested):
the stdin approach for prompt passing is solid and fixes the real Windows cmd.exe limit issue. the actual code changes look good. a few things though:

1. **PR description doesn't match the diff.** the title and body claim 3 independent fixes (stdin passing, Job Objects process tree cleanup, CancelIoEx cancellable stdin reads) but only the first one is implemented. `procgroup_windows.go` and `input_windows.go`/`input_unix.go` are not modified. the plan file in `docs/plans/completed/` is accurate, so pls update the PR description to match what's actually here

2. **linter failures** - two unused `//nolint:revive` directives at `executor_test.go:772` and `:786`. need to remove those

3. **`os.Truncate` vs `f.Truncate`** in `progress.go:195` - I get the Windows O_APPEND issue, but the existing comment at line 190 explicitly says "reads from the locked fd directly to avoid TOCTOU path-vs-inode mismatch". switching to path-based `os.Truncate(f.Name(), 0)` contradicts that safety model. low risk since the lock is held, but the comment should be updated to reflect the new tradeoff

4. **missing test for `codex-as-claude.sh`** - gemini and opencode wrappers both got `_test.sh` updates for stdin behavior, but codex wrapper has no test file despite identical changes

umputun (commented):
thx for the fixes. nolint directives removed, `progress.go` comment updated, codex wrapper test added - all good.

the one remaining thing is the PR description. it still claims 3 fixes (Job Objects, CancelIoEx) that aren't in the diff. the plan file is accurate, so just need to align the PR body with what's actually here. pls update and I'll approve

stanurkov:
> the one remaining thing is the PR description. it still claims 3 fixes (Job Objects, CancelIoEx) that aren't in the diff. the plan file is accurate, so just need to align the PR body with what's actually here. pls update and I'll approve

Sorry, didn't press the Update button before writing the reply. Fixed too

PR #274, opened 2026-04-09
feat: add file-backed plan requests for --plan (@file + @@literal)

DenisVanyushkin:
## Summary

- Add support for `--plan @path/to/request.md` to load plan requests from a file
- Support `@@literal` to preserve a literal leading `@` in inline requests
- Thread request-file metadata through plan-mode logging, progress filenames, and prompt templates
- Update docs and tests for both inline and file-backed plan request flows

## Motivation

Some plan requests are easier to write, review, and reuse as Markdown files rather than a single shell argument. File-backed input makes larger requests shell-safe and easier to maintain, while preserving the current inline `--plan "..."` workflow.

## Usage

```bash
# inline request
ralphex --plan "add user authentication"

# file-backed request
ralphex --plan @docs/requests/add-user-authentication.md

# literal leading @
ralphex --plan @@mention-based feature
```

## Test plan

- [x] `make test`
- [x] `make lint`
- [x] Request resolution tests cover inline text, `@file`, `@@literal`, missing files, blank files, and basename collisions
- [x] Plan-mode runner, prompt, and progress tests cover request-file logging and template substitution
- [x] Docs updated in `README.md`, `CLAUDE.md`, and `llms.txt`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -209,7 +209,7 @@ Key files:
 
 ### Plan Creation Mode
 
-The `--plan "description"` flag enables interactive plan creation:
+The `--plan` flag enables interactive plan creation from either inline text or `@path/to/request.md` input. Use `@@literal` if the request itself should start with `@`:
 
 - Claude explores codebase and asks clarifying questions
 - Questions use QUESTION signal with JSON: `{"question": "...", "options": [...]}`
--- a/README.md
+++ b/README.md
@@ -218,9 +218,10 @@ The `--plan` flag provides a simpler integrated experience:
 
 ```bash
 ralphex --plan "add health check endpoint"
+ralphex --plan @docs/requests/add-health-check.md
 ```
 
-Claude explores your codebase, asks clarifying questions via a terminal picker (fzf or numbered fallback), and generates a complete plan file in `docs/plans/`. When reviewing the draft, you can accept, revise with text feedback, open it in `$EDITOR` for interactive annotation, or reject it.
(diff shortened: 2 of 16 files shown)


umputun (commented):
thx for the PR, the implementation is clean and well-tested.

I have a concern about the overall approach though. the `@file` convention works, but it's not the most unix-idiomatic way to handle this. the standard pattern would be stdin support, i.e. `cat request.md | ralphex --plan -` or `ralphex --plan - < request.md`. this is more composable (pipes work with any tool), doesn't need the `@@` escape for literal `@`, and is familiar to any unix user. `gh` does this with `--body-file -`, for example.

the `@@` escape is the main symptom of the problem - users who write `--plan "@channel notification feature"` will get a cryptic "file not found" error with no hint about `@@`. with stdin, this edge case just doesn't exist.

`$(cat file.md)` also works today with zero changes for most cases, though it may have issues with very large files hitting shell argument limits.

I'd suggest going with the stdin approach instead.

also noticed a few implementation issues (dead code in `runPlanMode`, redundant error wrapping, missing hint about `@@` in error messages), but those are secondary to the approach question above.

DenisVanyushkin:
thx for the thoughtful review.

I agree with the overall point. the @file convention works, but stdin is the more unix-idiomatic interface here and it avoids the awkward @@ escape plus the confusing literal-@ edge case. the composability argument makes sense too.

given that, I don't think it makes sense to keep pushing on this version. I'll close this PR rather than iterate on an approach we already agree is the wrong one.

also thx for calling out the secondary implementation issues. those are useful notes as well, but I agree the interface choice is the main thing.

PR #281, opened 2026-04-14
fix: downgrade codex non-zero exit to warning when stdout has output

g-agent-lab:
## Problem

Codex CLI (v0.105+) returns exit code 1 via an internal `error_seen` flag when it encounters transient API errors (`ServerNotification::Error`, `TurnCompleted` with `Failed`/`Interrupted` status), even when the model has already generated a complete response with review findings in stdout.

Currently, any non-zero exit from codex is treated as a fatal error in `CodexExecutor.Run()`, which propagates to `runExternalReviewLoop()` and aborts the entire pipeline:

```
error: runner: codex loop: codex execution: codex exited with error: command wait: exit status 1
stderr: 1. High: the new policy layer is still not the authoritative gate...
```

The findings are visible in stderr (as context), but the review loop never reaches Claude evaluation — the usable output is discarded.

## Root Cause

In `codex-rs/exec/src/lib.rs`, codex sets `error_seen = true` on:
- `ServerNotification::Error` with `will_retry: false`
- `TurnCompleted { status: Failed }`
- `TurnCompleted { status: Interrupted }`

Then exits with `process::exit(1)` if `error_seen` is true. This can happen after the model already wrote its full response to stdout (e.g., a transient API error on the final turn acknowledgment).

## Fix

Add a post-pattern-matching downgrade in `CodexExecutor.Run()`: when codex exits non-zero but produced substantive stdout output and no error/limit patterns matched, clear the error and log a warning.

**Priority order (preserved):**

| Priority | Condition | Result |
|----------|-----------|--------|
| 1 | Pipe errors (stderr/stdout read) | Fatal error |
| 2 | Context cancellation | Fatal error |
| 3 | Exit code + limit pattern match | `LimitPatternError` (retry) |
| 4 | Exit code + error pattern match | `PatternMatchError` (fatal) |
| 5 | **Exit code + non-empty output + no patterns** | **Warning (continue)** ← new |
| 6 | Exit code + empty output | Fatal error |

The downgrade only applies to exit code errors (`isExitError` flag), not to pipe read failures, so `TestCodexExecutor_Run_ErrorPriority` (stderr error > wait error) is unaffected.

## Tests

- `TestCodexExecutor_Run_WaitErrorWithOutput` — exit 1 + findings → no error, warning logged
- `TestCodexExecutor_Run_WaitErrorWithOutputAndPattern` — exit 1 + rate limit output → pattern error takes priority
- `TestCodexExecutor_Run_WaitErrorNoOutput` — exit 1 + empty output → fatal error (unchanged)
- Updated 3 existing tests where exit 1 + unmatched output is now downgraded

All `./pkg/...` tests pass.

## Reproduction

1. Run a ralphex plan with codex review enabled
2. Codex generates findings but encounters a transient OpenAI API error on the last turn
3. Codex exits with code 1
4. **Before fix:** pipeline aborts with `codex execution: codex exited with error`
5. **After fix:** warning logged, findings passed to Claude evaluation normally

### DIFF
--- a/pkg/executor/codex.go
+++ b/pkg/executor/codex.go
@@ -164,6 +164,7 @@ func (e *CodexExecutor) Run(ctx context.Context, prompt string) Result {
 
 	// determine final error (prefer stderr/stdout errors over wait error)
 	var finalErr error
+	var isExitError bool // true when finalErr comes from non-zero exit code (not pipe errors)
 	switch {
 	case stderrRes.err != nil && !errors.Is(stderrRes.err, context.Canceled):
 		finalErr = stderrRes.err
@@ -173,6 +174,7 @@ func (e *CodexExecutor) Run(ctx context.Context, prompt string) Result {
 		if ctx.Err() != nil {
 			finalErr = fmt.Errorf("context error: %w", ctx.Err())
 		} else {
+			isExitError = true
 			// include stderr tail for error context when codex exits with non-zero status
 			if len(stderrRes.lastLines) > 0 {
 				finalErr = fmt.Errorf("codex exited with error: %w\nstderr: %s",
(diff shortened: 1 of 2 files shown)


umputun (commented):
the core assumption here - "non-empty stdout = complete output" - isn't verifiable. the codex-rs analysis in the PR description actually shows why: `ServerNotification::Error` and `TurnCompleted { status: Failed }` can fire at any point during generation, not only "on the final turn acknowledgment." nothing in codex-rs guarantees the model finished writing before `error_seen` is set. so we can't distinguish "complete response + late transient error" from "crashed mid-output."

feeding truncated output to claude is actively risky - partial "thinking" text or half-finished analysis could convince claude to make wrong fixes with high confidence. the current behavior (abort on non-zero exit) is the safe default.

the scenario you're solving for - codex hits a transient API error - is already covered by the configurable error/limit pattern mechanism. if there's a new error message that keeps appearing, the right fix is adding it to `codex_error_patterns` or `codex_limit_patterns` config so it gets proper handling (retry or clean exit), not silently swallowing exit codes.

I'm leaning toward rejecting this one, but if you think I'm missing smth - pls explain what guarantees that stdout is complete when codex exits non-zero.

PR #305, opened 2026-04-25
fix: resume watch-mode tailing after flock race marks session completed

### LINKED ISSUE #283: Watch mode: tailing stops after RefreshStates marks session completed (flock not held by executor)
## Description

When running ralphex in watch mode (`ralphex -s -w /path`) as a separate PM2 process alongside a task executor (`ralphex docs/plans/plan.md` without `-s`), the dashboard stops streaming logs mid-execution.

## Steps to reproduce

1. Start watch-mode dashboard as a persistent process:
   ```bash
   pm2 start ralphex --name ralph-web -- -s -w /home/user/dev
   ```

2. In a separate tmux session, start task execution without `-s`:
   ```bash
   cd ~/dev/my-project && ralphex docs/plans/my-plan.md
   ```

3. Open the dashboard in browser — initially shows progress and logs correctly.

4. After a few minutes (seems to coincide with `RefreshStates()` 5-second tick), the session is marked `completed` in the sidebar and **log streaming stops**.

5. The plan panel (right side) continues to update checkboxes correctly (reads file directly), but the central log area freezes.

6. The executor process is still running and writing to the progress file. New task iterations appear in the file but not in the dashboard.

## Root cause analysis

From reading the source code:

- `session_manager.go` `RefreshStates()` checks `IsActive()` every 5 seconds for tailing sessions.
- `IsActive()` uses `progress.TryLockFile()` (flock) to determine if another process holds the lock.
- When the executor runs **without `-s`**, it does not acquire an flock on the progress file.
- `RefreshStates()` successfully acquires the lock → concludes the session is not active → calls `session.SetState(SessionStateCompleted)` + `session.StopTailing()`.
- Once tailing stops, there is no mechanism to restart it when new writes are detected — `handleProgressFileChange()` calls `startTailingIfNeeded()` which checks `session.GetState() != SessionStateActive` and skips since state is `completed`.

## Expected behavior

Watch mode should continue streaming logs as long as the progress file is being written to, regardless of flock state. The flock-based detection makes sense for same-process `-s` mode but breaks the watch-mode use case where executor and dashboard are separate processes.

## Suggested fix

Option A: In watch mode, use file modification time instead of flock to determine activity. If the file was modified within the last N seconds, consider it active.

Option B: In `handleProgressFileChange()`, if a `Write` event arrives for a `completed` session, transition it back to `active` and restart tailing.

Option C: Have the executor (without `-s`) also acquire flock on the progress file.

## Environment

- ralphex v0.26.3-e1ea16f-20260408T225530
- Linux 6.8.0-94-generic (Ubuntu)
- PM2 for process management

umputun:
Fixes #283.

When the watch-mode dashboard runs as a separate process from the executor (`pm2 start ralphex -s -w /path` + separate `ralphex docs/plans/...`), `RefreshStates` would mark the session `completed` whenever the 5s tick hit a moment where `TryLockFile` won the race against the executor's flock. Once marked completed, `startTailingIfNeeded` refused to restart tailing, so the dashboard froze even though the executor was still writing to the progress file.

**Fix** — option B from the issue thread: on a write event for a `completed` session, reactivate it and resume tailing from a tracked offset. Flock-based detection in `RefreshStates` stays — reactivation is the recovery path, not a replacement.

**Mechanics**

- `Tailer.Offset()` and `Tailer.StartFromOffset(offset)` expose and seek the internal offset
- `Session.lastOffset` captured inside `StopTailing()` before the tailer is stopped
- `Session.Reactivate()` starts a new tailer at `lastOffset` (or `fromEnd` if offset is 0), then flips state to `active` only on success
- `loadProgressFileIntoSession` records bytes read pre-trim, so CRLF/LF/no-newline all count correctly
- `Watcher.handleProgressFileChange` calls `Reactivate()` for the specific written path when the session is `completed` and `IsLoaded()` (avoids racing the loader)

**Windows note** — `flock` is a no-op on Windows so `IsActive` always returns false. Reactivation-on-write becomes the *primary* streaming mechanism on Windows, not just a recovery path. No platform-specific code needed.

**Tests**
- TDD reproduction: `TestWatcher_ResumesStreamingAfterFlockRace` (red without the fix, green with it)
- offset/seek edges: zero-fallback, beyond-file-size, CRLF, LF, no-trailing-newline
- `Reactivate`: idempotent, resume-from-offset (no duplicates), failed-start-leaves-state-unchanged, on-closed-session
- watcher: only the written path is reactivated (sibling completed sessions stay completed)
- 86.9% coverage in `pkg/web`

Verified end-to-end against the toy project — dashboard streaming survives the 5s `RefreshStates` tick mid-run.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -67,6 +67,7 @@ docs/plans/         # plan files location
 - Streaming output with timestamps
 - Progress logging to files
 - Progress file locking (flock) for active session detection
+- Watch-mode dashboard reactivates completed sessions on fsnotify Write events, resuming tailing from the recorded `Session.lastOffset` — recovery path for the flock race in `RefreshStates` that can prematurely mark a still-running session as completed (issue #283). `Session.Reactivate()` is idempotent and scoped to the exact path that received the write; `loadProgressFileIntoSession` records `lastOffset` after the initial load so reactivation does not re-emit already-replayed events
 - Progress file fresh start: files ending in a `Completed:` footer are truncated on reuse; files ending in a `Failed:` footer (written when `Logger.SetFailed` was called before `Close`) or with no footer preserve existing content and write a `--- restarted at ... ---` separator, so retried failed/aborted runs keep history (issue #288). `SetFailed` is called in `cmd/ralphex/main.go` for `r.Run` errors (including `ErrUserAborted`), dashboard start errors, and any error return from `runWithWorktree`
 - Multiple execution modes: full, tasks-only, review-only, external-only/codex-only, plan creation
 - `--base-ref` flag overrides default branch for review diffs (branch name or commit hash)
--- a/docs/plans/20260424-watch-mode-reactivate-tailing.md
+++ b/docs/plans/20260424-watch-mode-reactivate-tailing.md
@@ -266,12 +266,12 @@ The feedEvents goroutine acquires its own `RLock` once running — no deadlock b
 
 ### Task 6: Verify acceptance criteria
 
-- [ ] re-read issue #283 and confirm Option B semantics match this implementation
-- [ ] run full test suite: `make test`
-- [ ] verify race-free: `go test -race ./pkg/web/...`
-- [ ] run linter: `make lint` (fix all issues — no "pre-existing" dismissals)
-- [ ] run formatter: `make fmt`
-- [ ] check coverage: `go test -cover ./pkg/web/...` — must be >=80% for touched files, ideally no regression
+- [x] re-read issue #283 and confirm Option B semantics match this implementation
+- [x] run full test suite: `make test`
+- [x] verify race-free: `go test -race ./pkg/web/...`
+- [x] run linter: `make lint` (fix all issues — no "pre-existing" dismissals)
+- [x] run formatter: `make fmt`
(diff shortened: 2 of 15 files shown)


umputun:
addressed findings 1 and 3 in 28f14bb. pushing back on finding 2.

**watcher_test.go:723** (active-session test) — valid. once `w.Start` runs `DiscoverRecursive`, `updateSession` sees `IsActive=false` (no flock holder), flips state to completed and stops the tailer, so the assertions pass for the wrong reason. fixed by adding a unix-only test helper that holds an exclusive flock on the progress file before the watcher starts. windows skips since flock-based detection is a no-op there anyway.

**session.go:303-306** (startTailerLocked godoc) — valid, doc was stale. rewrote to describe the current implementation: `feedEvents` receives tailer/stopCh/feedDone as args at spawn time and doesn't acquire `s.mu`.

**watcher_test.go:547** (reactivation test) — disagree, the test is fine. `Discover` registers the session in `m.sessions` *after* `updateSession` returns, and `updateSession` runs `loadProgressFileIntoSession` synchronously inside the `MarkLoadedIfNot` block (session_manager.go:213). so `sm.Get(sessionID) != nil` ⟹ `updateSession` completed ⟹ `IsLoaded() == true`. the test's existing wait condition is sufficient — there's no observable window where `Get` returns the session before it's loaded.

Issue #306, opened 2026-04-27
Support OpenSpec change directories as plan input


bronislav:
# Support OpenSpec change directories as plan input

I use ralphex daily, but the company I work at is pushing the OpenSpec workflow and I need to align with it — ideally without dropping ralphex. Hence this request.

## Problem

[OpenSpec](https://openspec.dev/) is a spec-driven workflow where change proposals live in `openspec/changes/<id>/` with a `tasks.md` (checkbox task list), `proposal.md`, `design.md`, and `specs/` delta files. Pointing ralphex at `openspec/changes/<id>/tasks.md` today doesn't work:

1. Parser at `pkg/plan/parse.go:46` accepts only `### Task N:` / `### Iteration N:` headers. OpenSpec uses `## 1. Phase 1` — checkboxes are silently dropped.
2. Checkboxes outside recognized task sections are ignored (`pkg/plan/parse.go:106-115`).
3. On success, `MovePlanToCompleted` runs unconditionally for `ModeFull`/`ModeTasksOnly` (`cmd/ralphex/main.go:541`). For OpenSpec, the directory must stay put — archiving is a human step via `openspec archive`.
4. Sibling context files (`proposal.md`, `design.md`, spec deltas) aren't fed to Claude, so tasks lose their "why".

## Proposal

Detect OpenSpec by path. When the plan file matches `**/openspec/changes/<id>/tasks.md`:

- Accept `## N. <title>` headers as task sections (in addition to `### Task N:`)
- Skip the move-to-completed step
- Include sibling `proposal.md`, `design.md`, and `specs/**/*.md` in the prompt context so Claude sees the full change package

No new flags. One command works:

```bash
ralphex openspec/changes/add-dark-mode/tasks.md
```

## Non-goals

- Running `openspec archive` on completion — stays manual
- Producing OpenSpec changes from `--plan` mode
- Modifying specs or writing deltas

Tighter integration (archive hook, `--plan` → OpenSpec output) can follow if this proves useful.

## Alternatives considered

- **Generic flags** (`--no-archive`, `--context-glob`): more surface area, user has to remember them every run.
- **Wrapper script** that flattens an OpenSpec change into a ralphex plan: two moving parts, transient files, and still needs a `--no-archive` flag.

Path-triggered detection keeps the UX to one command and the code delta small.

umputun:
i don't want to bake openspec awareness into ralphex. it's a generic plan executor, coupling it to a specific spec workflow is the wrong direction.

[...]

PR #307, opened 2026-04-27
feat: add move_plan_on_completion config option

bronislav:
## Summary

Adds `move_plan_on_completion` config option (default `true`) that controls whether completed plans move to `docs/plans/completed/` on success. Set to `false` for workflows that manage plan lifecycle externally (e.g. spec-driven tooling with separate archive steps).

Part 1 of 2 for #306 (OpenSpec integration). Kept scope tight: no OpenSpec-specific detection here, just a generic toggle that any external workflow can use.

## Changes

- `pkg/config/values.go` — `MovePlanOnCompletion` + `MovePlanOnCompletionSet` fields, INI loader, merge block (local overrides global even when explicit `false`)
- `pkg/config/config.go` — `MovePlanOnCompletion` on `Config` struct, defaults to `true`
- `pkg/config/defaults/config` — commented template entry
- `cmd/ralphex/main.go` — `shouldMovePlan()` predicate guards the existing `MovePlanToCompleted` call; nil-safe against `req.Config`
- Tests: `values_test.go`, `config_test.go`, `main_test.go` (including nil_config case)
- Docs: `README.md`, `CLAUDE.md`, `llms.txt`

Default behavior is unchanged.

## Test plan

- [x] `make test` passes
- [x] `make lint` clean
- [ ] Reviewer: verify existing behavior preserved when option is unset (default `true`)
- [ ] Reviewer: verify `move_plan_on_completion = false` in local config takes precedence over global `true`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -290,6 +290,7 @@ GOOS=windows GOARCH=amd64 go build ./...
 - `wait_on_limit` config option: duration to wait before retrying on rate limit (e.g., "1h", "30m"). CLI flag `--wait` takes precedence. Disabled by default
 - `session_timeout` config option: per-session timeout for claude (e.g., "30m", "1h"). Kills hanging sessions and continues to next iteration. CLI flag `--session-timeout` takes precedence. Disabled by default
 - `idle_timeout` config option: kills claude sessions when no output for specified duration (e.g., "5m"). Resets on each output line, only fires when session goes silent. CLI flag `--idle-timeout` takes precedence. Disabled by default
+- `move_plan_on_completion` config option: controls whether completed plans move to `docs/plans/completed/` on success. Default `true`. Disable for workflows that manage plan lifecycle externally (spec-driven tooling with separate archive steps)
 
 ### Local Project Config (.ralphex/)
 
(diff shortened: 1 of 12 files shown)


umputun (commented):
few things worth fixing:

1. **`shouldMovePlan` nil-check exists only for tests** (`cmd/ralphex/main.go:828-830`). No production caller passes a nil `Config`. Every `executePlanRequest` literal assigns it, and `executePlan` already dereferences `req.Config` at line 540 without a nil check. The `req.Config != nil` clause is there only so `TestShouldMovePlan/nil_config` passes. Drop the nil branch and the test case.

2. **`MovePlanOnCompletionSet` on `Config` is dead state** (`pkg/config/config.go:77`). Production never reads it; only `config_test.go:387,432` assert it. The `*Set` field is needed on `Values` for merge logic, but on `Config` (post-assembly) it's unused. Other existing fields do the same (`CodexEnabledSet`, `FinalizeEnabledSet`, etc.) but no reason to perpetuate it for the new one. Drop the field on `Config` and the matching assertions.

3. **default-true split across layers** (`pkg/config/config.go:281-285`). Default is applied post-assembly in `loadConfigFromDirs` instead of in the embedded template. Mirrors `NotifyOnError`/`NotifyOnComplete` precedent so not strictly wrong, but cleaner to uncomment `move_plan_on_completion = true` in `pkg/config/defaults/config` and drop the post-assembly patch. Minor.

4. **redundant test assertions** (`pkg/config/config_test.go:412,431-432`). Dedicated `TestLoad_MovePlanOnCompletion` already covers default/explicit-true/explicit-false; `TestLoad_AllUserValues` doesn't need to re-assert the same field. Resolves automatically if you apply #2.

naming: `move_plan_on_completion` doesn't quite match the project's `_enabled` / `use_` convention (`codex_enabled`, `finalize_enabled`, `use_worktree`). I agreed to it in #306 so it's settled, just noting.

umputun (approved):
lgtm, thx

PR #309, opened 2026-04-28
feat: configurable task header patterns

bronislav:
## Summary

Adds `task_header_patterns` config option (and `{{TASK_HEADER_PATTERNS}}` template variable) so the plan parser can recognize task sections beyond the default `### Task {N}: {title}` / `### Iteration {N}: {title}` shapes. Useful for spec-driven workflows (e.g. OpenSpec) that use headers like `## {N}. {title}`.

- template-to-regex compiler in `pkg/plan/patterns.go` with `{N}` (required) and `{title}` (optional) placeholders
- config field + loader with embedded default, per-field fallback, CLI passthrough
- `ParsePlan`/`ParsePlanFile` accept variadic `patterns ...string`
- `task.txt` prompt rewritten to reference `{{TASK_HEADER_PATTERNS}}`
- completed plan archived under `docs/plans/completed/`

## Test plan

- [x] `make test`
- [x] `make lint`
- [x] e2e run with toy project using default patterns
- [x] e2e run with a custom pattern (e.g. `## {N}. {title}`)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -142,6 +142,10 @@ linters:
           - usestdlibvars
         path: _test\.go$
       - linters:
+          - gosec
+        text: "G704: SSRF via taint analysis"
+        path: _test\.go$
+      - linters:
           - errcheck
         text: "Error return value of .*.Close.*is not checked"
       - linters:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -291,6 +291,7 @@ GOOS=windows GOARCH=amd64 go build ./...
 - `session_timeout` config option: per-session timeout for claude (e.g., "30m", "1h"). Kills hanging sessions and continues to next iteration. CLI flag `--session-timeout` takes precedence. Disabled by default
 - `idle_timeout` config option: kills claude sessions when no output for specified duration (e.g., "5m"). Resets on each output line, only fires when session goes silent. CLI flag `--idle-timeout` takes precedence. Disabled by default
 - `move_plan_on_completion` config option: controls whether completed plans move to `docs/plans/completed/` on success. Default `true`. Disable for workflows that manage plan lifecycle externally (spec-driven tooling with separate archive steps)
+- `task_header_patterns` config option: comma-separated list of preset names (`default`, `openspec`) or raw Go regexes controlling which headers the plan parser recognizes as task sections. Capture group 1 = task id (required), capture group 2 = title (optional). Default: `default` (matches `### Task N: title` and `### Iteration N: title`). Use `openspec` or a raw regex for spec-driven workflows that use different header shapes. Resolved via `plan.ResolveHeaderPatterns` in `pkg/plan/presets.go`. Key files: `pkg/plan/presets.go` (preset registry, `ResolveHeaderPattern`, `ResolveHeaderPatterns`, `DefaultHeaderPatterns`, `PresetDescription`); `pkg/plan/parse.go` (`ParsePlan(content string, patterns []*regexp.Regexp)`, `ParsePlanFile(path string, patterns []*regexp.Regexp)` — callers resolve `[]string` config values via `plan.ResolveHeaderPatterns` before passing; `nil`/empty slice falls back to `DefaultHeaderPatterns()`)
 
 ### Local Project Config (.ralphex/)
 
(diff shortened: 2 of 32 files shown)


bronislav:
@umputun this is quite a big pull request and i am worried that I might miss some important things. I am going to mark it as ready for review even though I did not test it manually myself yet. I am going to do it over next several days.

I open to changes in the design or approach if you think they needed. I never used web dashboard myself to monitor several sessions, so my assumptions around that might be different from what you think is appropriate.

bronislav:
Already found a "bug": ralphex creates a branch and worktree based on the plan file name and ignoring directory it resides in. I probably need to create a separate pull request to make branch configurable or dependent on the subdirectory as well.

umputun (commented):
@bronislav I asked for configurable patterns in #306 but didn't expect a +2400 PR. The bulk isn't justified for the stated need. Here's what I think we should ship instead.

**Two big offenders:**

1. `pkg/plan/patterns.go` (208 LOC) + `patterns_test.go` (394 LOC). The `{N}/{title}` DSL with its own compiler, brace validator, whitespace normalizer. Our users are devs running ralphex on their own plans, they can write regex. I'd rather have:

```ini
task_header_patterns = ^### (?:Task|Iteration) ([^:]+?):\s*(.*)$, ^## (\d+(?:\.\d+)?)\.\s*(.*)$
```

Capture group 1 = task id (required), capture group 2 = title (optional). Compile via `regexp.Compile`, surface compile errors as-is. About 50 prod LOC instead of 208, far fewer tests.

Default stays exactly what we have today (Task/Iteration only, no behavior change for existing users). For common shapes we add a small named-preset registry in `pkg/plan` as a `map[string]string`. Seed it with two entries: `default` (current Task/Iteration regex, what `task_header_patterns` resolves to when unset) and `openspec` (OpenSpec header regex). The value of `task_header_patterns` is either a known preset name, looked up in the map, or a raw regex passed to `regexp.Compile`:

```ini
# opt into openspec parsing
task_header_patterns = openspec

# or write a regex for some other shape
task_header_patterns = ^# Phase (\d+):\s*(.*)$
```

OpenSpec users get a one-liner. Power users can write regex. Adding the next spec flavor later is a one-entry map change, not a config schema change.

2. Progress-log persistence + dashboard recovery. `TaskHeaderPatterns:` line in progress files, re-emission at restart markers, `ParseProgressHeader` full-file scan, `loadSessionPlanWithFallback`. This is a configuration transport bolted onto an execution log to handle "single dashboard process watches multiple repos with different non-default header configs." I don't think anyone is doing that. The dashboard reading from its own config plus parser defaults is fine. Drop the whole layer (around 250-400 LOC across progress/session/web plus tests).

**What I'd keep:**

- `closesTask` / `headingLevel` refactor in `parse.go`, replaces the hand-rolled `isH2` / `isH1AfterTitle` tangle, real readability win
- `task_header_patterns` config field with comma-separated list (regex instead of templates)
- prompt rewrite in `task.txt` to expose configured shapes via `{{TASK_HEADER_PATTERNS}}`
- README/CLAUDE.md doc updates

Net shape: roughly +200/-50 across 8 or so files, down from +2403/-177 across 30.

**Smaller things, whatever shape we ship:**

- `Config.TaskHeaderPatternsSet` and `CompileTaskHeaderPattern(s)` are exported with no external callers, lowercase them
- `ParsePlan(content, patterns ...string)` variadic doesn't pay rent, every production caller has `[]string` and spreads it. Just take `[]string`
- if any progress-log machinery survives, `ParseProgressHeader` reads the entire progress file on every fsnotify event, needs to be tail-bounded like `isProgressCompleted`

bronislav:
I completely agree with your feedback. Going to address it.

umputun (changes requested):
closer, but still needs changes:

1. PR conflicts with master. Pls rebase first.
2. progress/web transport is still here (`pkg/progress/*`, `pkg/web/session_progress.go`, `pkg/web/server.go`). I asked to drop this layer; dashboard can use its own config/defaults for now.
3. default preset regresses existing plan parsing. Old parser accepted flexible whitespace, new regex requires exact single spaces.
4. `openspec` should match the documented dotted form only. Current regex accepts `## 7 Title` and misparses `## 1.2 Title` as id `1`, title `2 Title`.
5. lint fails on the PR branch: 22 issues, including misspelled `defualt` in `pkg/plan/presets_test.go` and `nestif` in `pkg/config/config.go`.
6. remove the unrelated git-config discovery plan from this PR.
7. remove or narrow the broad `_test.go` G704 exclusion in `.golangci.yml`.
8. docs still contradict the new behavior in places, e.g. README says task headers must use `### Task N:` / `### Iteration N:`.

core parser/config direction is better now, but this is still too broad for the feature.

bronislav:
I did not have time to complete this change for a few past weeks, this is why it is still in draft. I will try to complete it asap. Thank you for proactive review, though.

umputun:
closing this for now, it's been stalled in draft since May with no further progress. No worries, the configurable-header idea from #306 still stands. When you have time to trim it down and rebase, reopen this or push a fresh PR and I'll pick the review back up.

PR #314, opened 2026-04-30
feat: add provider override CLI flags

korjavin:
This PR adds per-run CLI flags that override configured provider settings for Claude-compatible task/review execution and external review tooling. 

### Why do we do these changes?

1.  **Prevent Profile Bloat**: Avoids creating a mess with multiple configuration files or separate `RALPHEX_CONFIG_DIR` profiles for simple variations in provider setup. One-off tests or temporary switches can now be handled directly via flags instead of maintaining redundant configuration state.
2.  **Improved Developer Experience**: By allowing `--claude-command` and `--claude-args` to override config settings *before* dependency checks, developers can easily point `ralphex` to experimental wrapper scripts even if the default Claude installation is missing or broken.
3.  **Config-to-CLI Consistency**: Adding these flags brings the provider configuration into the same override pattern already used for other settings like `--task-model` or `--session-timeout`.

### What changed?

-   **New CLI Flags**: Added `--claude-command`, `--claude-args`, `--external-review-tool`, and `--custom-review-script`.
-   **Explicit Override Tracking**: Implemented `ArgsSet` logic in the executor to allow explicit empty overrides (e.g., `--claude-args=`) to clear default Claude flags.
-   **Pre-Dependency Application**: Adjusted the boot sequence in `main.go` to apply CLI overrides immediately after config loading but before verifying the Claude dependency.
-   **Alias Support**: Added hidden underscore aliases (e.g., `--claude_command`) to match configuration file field names.
-   **Documentation**:
    -   Updated `README.md` with a new "Options" table and usage examples.
    -   Enhanced `docs/custom-providers.md` with details on per-run provider switching.
    -   Updated `llms.txt` for AI-assisted context.
-   **Tests**: Added comprehensive unit tests in `main_test.go` covering flag parsing, conflict resolution, and the dependency injection regression.

### Example Usage

```bash
# Use a custom wrapper and clear default args for a single run
ralphex --claude-command=/path/to/wrapper.sh --claude-args= docs/plans/feature.md

# Use a custom review script without editing config
ralphex --external-review-tool=custom --custom-review-script=./my-review.sh docs/plans/feature.md
```


umputun (changes requested):
couple of things before this can merge. The core feature is reasonable: the empty-args fix is solving a real documented gap in `pkg/config/values.go:435` (`mergeFrom` skips empty values, so `claude_args =` in config falls back to the embedded default), and the boot reorder is needed for `--claude-command` to actually work against `checkClaudeDep`.

**underscore aliases have no precedent**

`--claude_command`, `--claude_args`, `--external_review_tool`, `--custom_review_script` are the only underscore-aliased flags in the CLI. About 35 long flags in `cmd/ralphex/main.go`, none of the others ship aliases despite the same hyphen-vs-underscore mismatch with their config keys (`--task-model` / `task_model`, `--review-patience` / `review_patience`, etc.). The aliases cost 4 alias fields + 4 `*Set` bools + `validateAliasConflict` + `cliStringOverride` + a test matrix nothing else needs. I'd prefer dropping them. If consistent typo tolerance is something you want across the whole CLI, that's a separate scope.

**`--external-review-tool=custom` silently flips `CodexEnabled=true`**

`cmd/ralphex/main.go:1322-1327`. This is real back-compat work for users with `codex_enabled = false` legacy config, which is fine, but it lives in the wrong layer. A CLI flag silently mutating an unrelated config field is surprising. Cleaner: have `Runner.externalReviewTool()` defer to an explicit `ExternalReviewTool` regardless of `CodexEnabled`, then drop the implicit flip from `applyCLIOverrides`. At minimum the flag description should mention the side-effect.

korjavin:
FIxed

umputun (approved):
both points addressed cleanly:

- aliases dropped in 8996ab0, only the canonical hyphen forms remain
- `codex_enabled` flip moved out of `applyCLIOverrides` into `Runner.externalReviewTool()` in 7c3e82a, which now defers to an explicit `ExternalReviewToolSet` over the legacy `codex_enabled=false` back-compat. Regression test locks the behavior.

tests pass, lint clean. Lgtm, thx

PR #315, opened 2026-04-30
feat: add --branch flag to override worktree/branch name

bronislav:
## Summary

Related to #306. Replaces #310.

The `--branch=NAME` flag lets users explicitly set the branch/worktree name instead of relying on auto-derivation from the plan filename.

## Motivation

PR #310 attempted to fix branch name derivation with a heuristic (generic filename detection + parent directory fallback). As @umputun pointed out in the review, this approach is too fragile for such a simple case:

- The heuristic only helps when the filename is in a hardcoded set (`tasks`, `plan`, `plans`, `index`, `readme`) — listing all possible generic names is impossible, and users with `spec.md`, `description.md`, etc. get nothing
- The `--worktree` mode was the only path where the heuristic actually changed anything; regular mode already works because `preparePlanBranch` returns early if already on a feature branch
- A flag is strictly more general: works for any layout, custom namespacing (`feat/foo`), capitalized variants, arbitrary filenames
- ~5 lines, no regex, no edge cases, self-documenting in `--help`

## Changes

- Add `--branch=NAME` CLI flag (overrides branch name derived from plan filename)
- Thread `BranchOverride` through `executePlanRequest`, `preparePlanBranch`, `CreateBranchForPlan`, `CreateWorktreeForPlan`
- Both the worktree path (`.ralphex/worktrees/<name>`) and the progress logger use the override when set
- Works for worktree mode, normal mode, and plan-continuation mode
- Added tests for `CreateWorktreeForPlan` and `CreateBranchForPlan` with override

## Usage

```bash
# OpenSpec layout: tasks.md → would derive "tasks" without --branch
ralphex --worktree --branch=add-dark-mode openspec/changes/add-dark-mode/tasks.md

# Custom namespacing
ralphex --branch=feat/user-auth docs/plans/auth.md

# Generic plan filenames
ralphex --branch=my-feature docs/plans/plan.md
```

## Test plan

- [x] `go test ./pkg/git/...` — new override test cases pass
- [x] `make test` — full suite green
- [ ] Manual: `ralphex --worktree --branch=my-branch docs/plans/tasks.md` creates worktree at `.ralphex/worktrees/my-branch` on branch `my-branch`
- [ ] Manual: without `--branch`, existing behavior unchanged

### DIFF
--- a/cmd/ralphex/main.go
+++ b/cmd/ralphex/main.go
@@ -46,6 +46,7 @@ type opts struct {
 	IdleTimeout           time.Duration `long:"idle-timeout" description:"kill claude session after no output for this duration (e.g. 5m, 10m)"`
 	SkipFinalize          bool          `long:"skip-finalize" description:"skip finalize step even if enabled in config"`
 	Worktree              bool          `long:"worktree" description:"run in isolated git worktree"`
+	Branch                string        `long:"branch" description:"override branch name for worktree/branch creation (default: derived from plan filename)"`
 	PlanDescription       string        `long:"plan" description:"create plan interactively (enter plan description)"`
 	Debug                 bool          `short:"d" long:"debug" description:"enable debug logging"`
 	NoColor               bool          `long:"no-color" description:"disable color output"`
@@ -133,9 +134,10 @@ type executePlanRequest struct {
 	DefaultBranch string // actual default branch for branch/worktree creation (config or auto-detect)
 	BaseRef       string // base reference for review diffs and templates (--base-ref override or DefaultBranch)
 	NotifySvc     *notify.Service
-	WtCleanup     *worktreeCleanupFn  // worktree cleanup for interrupt handler; nil when not in worktree mode
-	ProgressLog   *progress.Logger    // pre-created logger (worktree mode); nil in normal mode
-	PhaseHolder   *status.PhaseHolder // pre-created holder (worktree mode); nil in normal mode
+	BranchOverride string             // branch name override (--branch flag); empty = derive from plan filename
+	WtCleanup      *worktreeCleanupFn  // worktree cleanup for interrupt handler; nil when not in worktree mode
+	ProgressLog    *progress.Logger    // pre-created logger (worktree mode); nil in normal mode
+	PhaseHolder    *status.PhaseHolder // pre-created holder (worktree mode); nil in normal mode
 }
 
 // worktreeCleanupFn holds a worktree cleanup function with mutex for safe cross-goroutine access.
@@ -301,14 +303,15 @@ func run(ctx context.Context, o opts) error {
 	}
 
 	return selectAndExecutePlan(ctx, o, executePlanRequest{
-		Mode:          mode,
-		GitSvc:        gitSvc,
-		Config:        cfg,
-		Colors:        colors,
-		DefaultBranch: defaultBranch,
-		BaseRef:       baseRef,
-		NotifySvc:     notifySvc,
-		WtCleanup:     wtCleanup,
+		Mode:           mode,
+		GitSvc:         gitSvc,
+		Config:         cfg,
+		Colors:         colors,
+		DefaultBranch:  defaultBranch,
+		BaseRef:        baseRef,
+		NotifySvc:      notifySvc,
+		WtCleanup:      wtCleanup,
+		BranchOverride: o.Branch,
 	}, selector)
 }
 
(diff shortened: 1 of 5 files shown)


cmd/ralphex/main.go
```diff
@@ -696,21 +702,7 @@ func runWithWorktree(ctx context.Context, o opts, req executePlanRequest) (err e
...
-			resolvedPlan = resolved
-		}
-		if rel, relErr := filepath.Rel(req.GitSvc.Root(), resolvedPlan); relErr == nil {
-			abs, absErr := filepath.Abs(rel) // resolve relative to CWD (now the worktree)
-			if absErr == nil {
-				wtPlanFile = abs
-			}
-		}
-	}
+	wtPlanFile := resolveWorktreePlanFile(req.PlanFile, req.GitSvc.Root())
```
bronislav, line 703: this code extracted to lower cyclomatic complexity warning raised by linter

umputun (changes requested):
couple of things to address before merge:

**1. `--branch` is silently dropped in `--plan` mode**

at `cmd/ralphex/main.go:292-301`, the `runPlanMode` branch builds `executePlanRequest` without `BranchOverride: o.Branch`. The non-plan path at line 314 sets it correctly. So `ralphex --plan "..." --branch=foo` ignores the override entirely, which is exactly the generated/generic-filename case the flag is supposed to help with. Fix is one line, but worth a test that locks the wiring.

**2. `CommitPlanFile` commit message ignores the override**

`pkg/git/service.go:330` still does `branchName := plan.ExtractBranchName(planFile)` to build the commit subject. With `--branch=my-feature` on a `2026-04-30-some-name.md` plan, the worktree is on `my-feature` but the auto-commit reads `add plan: some-name`. Copilot flagged this too. Easiest fix: have `CommitPlanFile` use the current branch from git rather than re-deriving from the filename, then this whole class of bug goes away.

**3. branch-name derivation is duplicated in 4 places**

`preparePlanBranch:189`, `CreateWorktreeForPlan` early at `:280`, `runWithWorktree` logger at `cmd/ralphex/main.go:650`, and the broken `CommitPlanFile:330`. Each one repeats `if branchOverride != "" { use it } else { ExtractBranchName(planFile) }`. Since most call sites are inside `Service`, makes sense as a method on `*Service`, something like `s.effectiveBranchName(planFile, override)`, then all four sites delegate to it. That alone would have prevented #2 and removes the risk of forgetting a site for #1.

**4. new override subtests are misplaced**

at `pkg/git/service_test.go:1261` and `:1281`. Both live inside `TestService_RemoveWorktree` (starts at 1200) but exercise `CreateWorktreeForPlan` and `CreateBranchForPlan`. Should move to `TestService_CreateWorktreeForPlan` (line 845) and `TestService_CreateBranchForPlan` (line 186) respectively. Tests-as-documentation gets weaker when the subtest name doesn't match its parent.

**5. gofmt drift**

`gofmt -d cmd/ralphex/main.go` shows a struct alignment diff on the `BranchOverride` field, the name is longer than the existing fields and the rest of the struct wants realignment. Golangci-lint doesn't catch it because gofmt isn't enabled in `formatters:` in `.golangci.yml`. `make fmt` fixes it. Small but a follow-up format pass will show a noisy diff if left.

**Notes that aren't blockers**

- Copilot's path-traversal flag on `service.go:282` is a false positive on impact. `filepath.Join` calls `Clean` which collapses `..`, then `git worktree add -b` rejects the bad ref before any directory is created (verified with `git check-ref-format --branch '../../outside'` exit 128). Cleanup never reaches an attacker-chosen path because `WtCleanup` is registered after successful creation. You may still want a defense-in-depth `sanitizeBranchName` for clearer error messages, but it's not a security issue.
- PR body says "Closes part of #306", just confirming you don't intend the merge to auto-close #306, since the issue has more aspects (header parsing, MovePlanToCompleted skip, sibling spec files) that this PR correctly leaves for follow-up. "Related to #306" might be safer phrasing.
- Consider also a test for `--branch=existing-feature-branch` (does it switch or error?) and one for the `runWithWorktree → CommitPlanFile` path with override set, which would have caught #2.

umputun (approved):
thx for working through these.

1, 2, 3, 5 all good. 2's `currentBranch()` approach is cleaner than what I suggested, sidesteps the threading entirely.

minor on 4: the `CreateBranchForPlan` override subtest is in the right place (`TestService_CreateBranchForPlan:427`), but the `CreateWorktreeForPlan` one ended up in `TestService_CommitPlanFile:1217` instead of `TestService_CreateWorktreeForPlan` - the body only exercises `CreateWorktreeForPlan`, no commit verification. One-line move would finish it.

also right call reverting f5cd336, the path-traversal validation wasn't needed.

lgtm with the test-placement nit, fine to merge as-is.

PR #316, opened 2026-04-30
fix: progress log filename ignores --branch override

bronislav:
## Summary

- Discovered during manual testing of #315: when `--branch` is used with a generic plan file (e.g. `tasks.md`), the progress log was still named after the plan file instead of the branch override
- Added `BranchOverride` field to `progress.Config` so `progressFilename` uses the explicit branch name as the filename stem when set
- Extracted `filenameWithStem` helper to eliminate duplication between the two switch blocks
- Added test cases covering `--branch` override for full, review, and codex-only modes

## Test plan

- [x] `make test` passes
- [x] `ralphex --worktree --branch=my-feature docs/plans/tasks.md` creates `.ralphex/progress/progress-my-feature.txt` instead of `progress-tasks.txt`

### DIFF
--- a/cmd/ralphex/main.go
+++ b/cmd/ralphex/main.go
@@ -417,10 +417,11 @@ func setupProgressLogger(o opts, req executePlanRequest, branch string) (progres
 	} else {
 		var err error
 		baseLog, err = progress.NewLogger(progress.Config{
-			PlanFile: req.PlanFile,
-			Mode:     string(req.Mode),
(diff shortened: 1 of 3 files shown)


umputun (approved):
lgtm, thx. Clean follow-up - keeping `planFile` as primary stem and only swapping in `branchOverride` when explicit was the right call, preserves the date-prefix `progress-2024-01-15-foo.txt` shape for normal use.

PR #349, opened 2026-05-17
feat(docker): add python development image

andreygaag:
Adds and publishes the ghcr.io/umputun/ralphex-python image with Python tooling (uv, ruff, mypy, pytest). Includes multi-arch build/merge jobs in docker.yml, a docker-build-python make target, and documentation updates in README and llms.txt.

## Summary by CodeRabbit

* **New Features**
  * Released a new Python development Docker image with pre-installed development tools and multi-architecture support for x86_64 and ARM64 platforms

* **Documentation**
  * Updated documentation with instructions for using the Python development image

[Review Change Stack](https://app.coderabbit.ai/change-stack/umputun/ralphex/pull/349?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,6 +10,7 @@ env:
   REGISTRY: ghcr.io
   IMAGE_BASE: ${{ github.actor }}/ralphex
   IMAGE_GO: ${{ github.actor }}/ralphex-go
+  IMAGE_PYTHON: ${{ github.actor }}/ralphex-python
 
 jobs:
   # Build base image for both architectures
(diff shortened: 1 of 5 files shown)


umputun (commented):
i don't think we should add this image.

Go image works as a reasonably universal reference: one Go version, `go test`, and a few common global tools. Python doesn't have the same shape. Real Python projects usually need their own Python version, native deps, lockfile/project deps, pytest plugins, and sometimes poetry/pdm instead of uv. A global `uv + ruff + mypy + pytest` image looks more like a template than a reusable runtime.

We already have documentation showing how to make a custom image, with the Go image as the example. Adding another template-like image doesn't improve that. It adds CI/release/GHCR surface and makes `ralphex-python` look like a supported artifact, while most Python users will still need to customize it.

There is also a concrete runtime bug in the current Dockerfile: `uv tool install` runs as root and leaves tool envs under `/root/.local/share/uv/tools`, while the container runs as `app`. The image builds, but `ruff` fails with permission denied in a normal container run. CodeRabbit's `latest` tag and unpinned-tool comments are valid too.

Declining this PR as an official Python image.

andreygaag:
Fair point.

PR #358, opened 2026-05-23
feat(config): configurable base directory for engine-created worktrees

OmniJeff:
## Motivation

`CreateWorktreeForPlan` hardcodes the worktree base directory to
`.ralphex/worktrees/<branch>` under the repo root. Operators who run ralphex
under `use_worktree = true` cannot redirect those worktrees elsewhere — for
example to a sibling directory next to the main checkout (which keeps the
worktrees visible to `git worktree list` from the main repo but out of `git
status`), or to a faster filesystem.

This PR adds a `worktree_path` config option to override that base directory.

## Behavior

- Default value: `.ralphex/worktrees` (matches the previous hardcoded path —
  the change is non-breaking).
- Relative paths are resolved against the repo root.
- Absolute paths are used as-is, so operators can place engine-created
  worktrees outside the repo entirely (e.g. `/var/lib/ralphex-worktrees` or
  `../ralphex-worktrees`).
- The per-plan directory is always `<worktree_path>/<branch>`.
- The config is wired into `git.Service` via a new `SetWorktreePath` method,
  mirroring the existing `SetCommitTrailer` pattern.

## Note on `.ralphex/.gitignore`

`EnsureLocalGitignore` still writes `progress/` and `worktrees/` so the
default path is covered. When operators set `worktree_path` to a location
inside the repo but outside `.ralphex/`, they are responsible for adding it
to their own `.gitignore` (typical for absolute or sibling paths, which is
the main use case). Documented inline in `defaults/config`.

## Tests

- `pkg/git/service_test.go` — `TestService_SetWorktreePath` covers default,
  relative, absolute, and reset-to-default paths.
- `pkg/config/values_test.go` — `TestValuesLoader_Load_WorktreePath` covers
  parsing and local-overrides-global precedence.

`go test ./...` passes the new tests; one pre-existing failure
(`TestService_CreateWorktreeForPlan/fails_when_branch_is_checked_out_in_another_worktree`)
exists on `master` too — it depends on the local git binary emitting "already
used by worktree" rather than "already checked out". Not touched by this PR.

`golangci-lint run ./...` reports 0 issues.

## Open to discussion

Happy to adjust naming (`worktree_path` vs `worktree_base_dir`), default
behavior, or the `Service` API shape (`SetWorktreePath` vs constructor
parameter) before merging. The current shape was chosen to keep the change
non-breaking and mirror existing patterns in the codebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### New Features
+
+- Configurable base directory for engine-created worktrees via the `worktree_path`
+  config option. Defaults to `.ralphex/worktrees` (relative to the repo root) so
+  the change is non-breaking. Absolute paths are supported for placing worktrees
+  outside the repo (e.g., `/var/lib/ralphex-worktrees` or a sibling directory
+  next to the main checkout).
+
 ## v1.3.1 - 2026-05-22
 
 ### New Features
--- a/cmd/ralphex/main.go
+++ b/cmd/ralphex/main.go
@@ -304,6 +304,7 @@ func run(ctx context.Context, o opts) error {
 		return fmt.Errorf("open git repo: %w", err)
 	}
 	gitSvc.SetCommitTrailer(cfg.CommitTrailer)
+	gitSvc.SetWorktreePath(cfg.WorktreePath)
 
 	// ensure repository has commits (prompts to create initial commit if empty)
 	if ensureErr := ensureRepoHasCommits(ctx, gitSvc, os.Stdin, os.Stdout); ensureErr != nil {
@@ -754,6 +755,7 @@ func runWithWorktree(ctx context.Context, o opts, req executePlanRequest) (err e
 		return fmt.Errorf("open worktree git service: %w", err)
 	}
 	wtGitSvc.SetCommitTrailer(req.Config.CommitTrailer)
+	wtGitSvc.SetWorktreePath(req.Config.WorktreePath)
 
 	// resolve plan file path inside the worktree so Claude operates on the local copy,
 	// not the original in the main repo. the plan was copied by CreateWorktreeForPlan.
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -94,8 +94,9 @@ type Config struct {
 
 	MovePlanOnCompletion bool `json:"move_plan_on_completion"`
 
-	WorktreeEnabled    bool `json:"worktree_enabled"`
-	WorktreeEnabledSet bool `json:"-"` // tracks if use_worktree was explicitly set in config
+	WorktreeEnabled    bool   `json:"worktree_enabled"`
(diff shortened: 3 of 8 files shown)


umputun (changes requested):
a few things to address before merge:

1. **in-repo non-default `worktree_path` isn't gitignored**. `EnsureLocalGitignore` still writes a fixed `.gitignore\nprogress/\nworktrees/\n` at `pkg/git/service.go:623`. If an operator sets `worktree_path = some-dir` inside the repo but outside `.ralphex/`, the worktree directory is not ignored, `hasChangesOtherThan(planFile)` trips on subsequent runs, and `git add -A` can commit it as a gitlink. Delegating this to operator-managed `.gitignore` is too sharp an edge. Please auto-add the resolved in-repo relative path to `.git/info/exclude` before `git worktree add` (local, non-committed, doesn't touch tracked `.gitignore`). Add a test that `worktree_path = tmp/wt` leaves `git status` clean after creation.

2. **rename `worktree_path` to `worktrees_dir`**. Matches the existing `plans_dir` convention. The plural matches the per-plan subdirectory layout already documented as `.ralphex/worktrees`.

3. **add `--worktrees-dir` CLI flag**. PR body says "CLI/INI/embedded default" but only INI is wired. Peer options (`commit_trailer`, `task_model`, `wait_on_limit`, `session_timeout`, `idle_timeout`, `move_plan_on_completion`, `preserve_anthropic_api_key`) all have CLI flags.

4. **dead setter at `cmd/ralphex/main.go:758`**. `wtGitSvc.SetWorktreePath(req.Config.WorktreePath)` runs on the git service inside the worktree, which never calls `CreateWorktreeForPlan`. Drop it. Unlike `SetCommitTrailer` on the same service (which is used by `wtGitSvc.CommitPlanFile`), this setter has no readers.

5. **drop the CHANGELOG hunk**. Project convention is `CHANGELOG: Never modify during development - updates are part of release process only`. I'll update it at release time.

6. **docs update**. `llms.txt` "Customization" section lists every other config option, please add `worktrees_dir`. CLAUDE.md "Worktree Isolation Mode" still describes `.ralphex/worktrees/<branch-name>` as the location, mention the override.

minor (take or leave):

- `strings.TrimSpace` is applied to the new key in `pkg/config/values.go:373` but not to other string keys parsed nearby (`task_model`, `review_model`, `claude_command`, etc.). Either normalize all or none.
- `TestService_SetWorktreePath` and `TestValuesLoader_Load_WorktreePath` are natural table-driven candidates per project convention.

branch is currently CONFLICTING with master. Master has a processor runner refactor (#364) that touches `cmd/ralphex/main.go`, so you'll need a rebase after these changes.

umputun:
closing this out, it's sat since the May change request with no updates. No problem, the idea's reasonable, it just needs the gitignore handling for a non-default `worktree_path` sorted. If you want to pick it back up some day, reopen or push a fresh PR and I'll take another look.

PR #359, opened 2026-05-23
feat(config): require_worktree opt-in for worktree-first workflows

OmniJeff:
## Motivation

Operators who manage worktrees externally (e.g. via a separate worktree
command that creates `../task-X` on a feature branch) currently get no
warning if they launch ralphex from the wrong place. The engine happily
creates a feature branch in the main checkout and starts committing there.

If the operator's review/ship workflow depends on the work landing in a
specific external worktree, the wrong-place launch corrupts both the main
checkout state and the operator's mental model of where the work lives.

## Behavior

This PR adds a `require_worktree` config option (default `false`). When
enabled, ralphex refuses to run from the main checkout's default branch
without `--worktree` and surfaces the missing worktree as a clear error
before any agent is invoked.

The error message points the operator at the two ways to satisfy the
requirement (create an external worktree and rerun, or pass `--worktree`
to let ralphex create one) and at the `worktree_path` config option as a
related knob (when the operator wants ralphex to create the worktree but
not at the default location).

## Non-breaking

- Default is `false`; existing behavior is preserved when the option is
  unset.
- On feature branches (the typical "launch from a pre-existing worktree"
  case) the check is satisfied and ralphex proceeds normally.
- The check fires before `EnsureLocalGitignore`, so an aborted run does not
  leave `.ralphex/.gitignore` in the main checkout as a side effect.

## Tests

- `cmd/ralphex/main_test.go` — `TestEnforceRequireWorktree` covers the
  branch-position logic (feature branch passes, default branch fails,
  empty-default fallback fails on `main`/`master`).
- `cmd/ralphex/main_test.go` — `TestSelectAndExecutePlan_RequireWorktree`
  verifies the abort happens before any state is written
  (`.ralphex/.gitignore` does not appear, no feature branch is created).
- `pkg/config/values_test.go` — `TestValuesLoader_Load_RequireWorktree`
  and `TestValues_mergeFrom_RequireWorktree` cover parsing, the
  `*Set` tracker, local-overrides-global precedence, and invalid-value
  rejection.

`go test ./pkg/config/... ./cmd/ralphex/...` passes the new tests;
pre-existing failures unrelated to this PR (`TestAutoPlanModeDetection`
subtests that require `codex` on `PATH`) are present on `master` as well.

`golangci-lint run ./...` reports 0 issues.

## Open to discussion

Naming, error-message wording, and the placement of the check
(`selectAndExecutePlan` vs earlier in `run`) are all up for review. The
current shape keeps the check close to the decision point (right before
the in-place `CreateBranchForPlan` call) so the guard's intent reads
clearly from the code path.

The check is intentionally limited to "default branch without `--worktree`"
because vanilla ralphex on a feature branch is a no-op for branch creation
— there's nothing for `require_worktree` to gate. Happy to expand the
guard to other cases if there's a real failure mode I missed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## Unreleased
+
+### New Features
+
+- Added a `require_worktree` config option. When enabled, ralphex refuses to run
+  from the main checkout's default branch without `--worktree` and surfaces the
+  missing worktree as a clear error before any agent is invoked. Operators who
+  manage worktrees externally (e.g. `git worktree add ../task-X -b task-X`) get
+  fail-fast feedback instead of an unintended feature-branch commit in the main
+  checkout. Default is `false`; the in-place flow on feature branches is
+  untouched.
+
 ## v1.3.1 - 2026-05-22
 
 ### New Features
--- a/cmd/ralphex/main.go
+++ b/cmd/ralphex/main.go
@@ -370,6 +370,17 @@ func selectAndExecutePlan(ctx context.Context, o opts, req executePlanRequest, s
 		return runWithWorktree(ctx, o, req)
 	}
 
+	// require_worktree safety: when the operator opts into a worktree-first workflow,
+	// refuse to create a feature branch in the main checkout. fires only when a branch
+	// would otherwise be created (mode requires it AND we are on the default branch).
+	// runs before EnsureLocalGitignore so the .ralphex/.gitignore is not written into
+	// the main checkout as a side effect of an aborted run.
+	if req.Config.RequireWorktree && planFile != "" && modeRequiresBranch(req.Mode) {
+		if rwErr := enforceRequireWorktree(req); rwErr != nil {
+			return rwErr
+		}
+	}
+
 	if err := req.GitSvc.EnsureLocalGitignore(); err != nil {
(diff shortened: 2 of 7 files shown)


umputun (changes requested):
needs changes before this is reviewable:

1. PR conflicts with master. Pls rebase first.
2. `cmd/ralphex/main.go:414` mentions `worktree_path`, but that is not in this PR or current `master`; looks like it belongs to separate PR #358. Remove the hint here, or make this PR explicitly depend on #358.
3. current text says this fails before any agent invocation, but the implementation only guards existing-plan full/tasks-only branch creation. Pls narrow the wording to that behavior and keep this as a branch-creation guard, not a general worktree requirement.
4. `--plan`, auto-plan, and plan continuation bypass the current guard. If the scope stays narrow, docs/error text should not claim those paths are covered.
5. remove the `CHANGELOG.md` change. Changelog updates are release-process only here.
6. config docs need to be updated where users look (`README.md` / `llms.txt`), not only in the defaults config template.
7. minor cleanup: reuse existing `branchExists` test helper instead of adding `gitHasBranch`.

umputun:
same story as #358, closing after the stall since May. It needs a rebase onto master and the `worktree_path` hint sorted out (it depends on #358). Happy to look again whenever you want to resume, just reopen or resubmit.

PR #372, opened 2026-05-28
fix(opencode): pass model and effort flags

mschedrin:
## Summary
- Parse Claude-compatible `--model` and `--effort` flags in `opencode-as-claude.sh`
- Map ralphex effort/reasoning values to OpenCode's supported `--variant` flag
- Add env var support for OpenCode variants/reasoning:
  - `OPENCODE_VARIANT`
  - `OPENCODE_EFFORT`
  - `OPENCODE_REASONING`
  - `OPENCODE_REVIEW_VARIANT`
  - `OPENCODE_REVIEW_EFFORT`
- Update `opencode-review.sh` to use `opencode run --model ... --variant ...` instead of the ineffective `agent.coder.reasoningEffort` config path
- Add/update wrapper tests and documentation
## Verification
- `bash scripts/opencode/opencode-as-claude_test.sh`
- `bash scripts/opencode/opencode-review_test.sh`
- `bash -n scripts/opencode/opencode-as-claude.sh && bash -n scripts/opencode/opencode-review.sh`
- Manual wrapper runs verified OpenCode persisted `variant=high` in its session DB

### DIFF
--- a/docs/custom-providers.md
+++ b/docs/custom-providers.md
@@ -230,6 +230,9 @@ claude_command = /path/to/scripts/opencode/opencode-as-claude.sh
 | Variable | Default | Description |
 |---|---|---|
 | `OPENCODE_MODEL` | (opencode default) | Model in provider/model format, e.g. `github-copilot/claude-opus-4.6` |
+| `OPENCODE_VARIANT` | (opencode default) | Model variant/reasoning effort, e.g. `high`, `medium`, or `low` |
+| `OPENCODE_EFFORT` | (opencode default) | Alias for `OPENCODE_VARIANT` when `OPENCODE_VARIANT` is unset |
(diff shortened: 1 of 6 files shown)


umputun (commented):
couple small things:

1. pls update `docs/custom-providers.md` too. It has the OpenCode wrapper env table and still only lists `OPENCODE_MODEL`, `OPENCODE_VERBOSE`, and `OPENCODE_CONFIG_CONTENT`, so the main provider doc will be stale.
2. minor/non-blocking: the new flag parser can accept missing values and consume the next flag as the value, e.g. `--model --effort high`. Might be worth rejecting empty or dash-prefixed values while touching this code.

otherwise looks good.

mschedrin:
Both comments addressed.

umputun (approved):
lgtm, thx

PR #378, opened 2026-06-09
Add pi provider support (wrapper + skills)

olomix:
## Summary

- Add `scripts/pi-as-claude/pi-as-claude.sh`, a wrapper that drives `pi --mode json --print` and translates its JSONL event stream into Claude `stream-json` (assistant `text_delta` → `content_block_delta`, `turn_end`/`agent_end` → `result`, tool events skipped unless `PI_VERBOSE=1`), so pi can be used as a `claude_command` drop-in for task and review phases
- Buffer pi token deltas into whole lines so `<<<RALPHEX:...>>>` signals survive translation even when split across deltas
- Support `PI_PROVIDER`/`PI_MODEL`/`PI_THINKING`/`PI_VERBOSE` env vars and map ralphex `--model`/`--effort` to pi `--model`/`--thinking` (`max` → `xhigh` with a note, since pi has no `max`); add `PI_EXTRA_ARGS` passthrough for auto-approving tools in non-interactive runs
- Add a pi-appropriate review-prompt adapter (sequential per-agent review, since pi has no parallel sub-agents)
- Add pi-adapted versions of all four ralphex skills under a new `assets/pi/` tree (`ralphex`, `ralphex-plan`, `ralphex-update`, `ralphex-adopt`), mapping Claude-only tooling (`Task`/`Explore`, `AskUserQuestion`, `run_in_background`/`TaskOutput`) to pi equivalents, plus a frontmatter validator
- Wrapper README + `docs/custom-providers.md` pi section; cross-reference pi in `README.md`, `llms.txt`, and `CLAUDE.md`
- Shell test suites for the wrapper (`pi-as-claude_test.sh`, `pi-as-claude_docs_test.sh`) and the pi skills (`assets/pi/skills_test.sh`)


umputun (changes requested):
the wrapper's the keeper here - clean clone of the existing `*-as-claude` pattern, tests pass. Two asks before it goes in.

**1. split out the pi skill tree**

can you drop `assets/pi/` and keep just the wrapper (`scripts/pi-as-claude/` + the one-line provider bullets in README/llms.txt/CLAUDE.md)? A few reasons:
- every other provider (codex, copilot, gemini, agy, opencode) is wrapper-only. `assets/pi/` would be the repo's first forked skill tree.
- it's ~1k lines hand-copied from `assets/claude/skills/` with no sync mechanism - no Makefile target, CI check, or test compares the two. The claude skills change roughly monthly with behavioral plan-format edits (loop prevention, date format, checkbox rules), so the pi copies would silently go stale. First signal would be a pi user hitting the old behavior.
- that also drops the `## pi Integration` H2 sections in README/llms.txt. They put pi level with Claude Code itself, out of proportion with the other providers.

not a no to pi-from-inside-pi skills forever, they just need a sync story (or real demand) first. Can revisit separately.

**2. prompt goes back onto pi's argv**

`pi-as-claude.sh:105` appends `$prompt` as a positional arg, then `:134` runs `pi "${pi_args[@]}"`. The script reads the prompt from stdin to dodge the cmd-line limit (per the header comment) but then puts it right back on the child's command line. A large review prompt with a full diff can blow past Linux's 128KB per-arg cap and fail before pi runs. Codex pipes via stdin, copilot writes a temp file - if `pi --print` reads from stdin do that, else a temp file works. `gemini-as-claude` has the same issue so it's not a regression, but worth fixing here.

not blocking, just flagging: stderr lines get re-emitted as `content_block_delta` (`:186-191`) and run through signal detection, so a stderr line with a literal `<<<RALPHEX:...>>>` could be read as a real signal. Low odds, and gemini does the same, so leaving it.

olomix:
let it be draft for some time, I'll check with some tasks before review.

umputun:
you took this out of draft but didn't flag it for re-review, and your last note said you wanted to keep it draft while testing some tasks. Ready for another look now, or still poking at it?

fwiw the two earlier asks look done - skill tree dropped, prompt goes to pi via stdin.

olomix:
Screenshot 2026-06-25 at 07 09 30

Hm... sorry, I thought I did ask for review. 

Yes, pi integration works for me. Though I don't use it too much, as claude revert -p plans, but the PR is working fine for me.

umputun (approved):
traced both asks, confirmed. Prompt's off argv now, written to a temp file and fed via stdin redirect (`pi "${pi_args[@]}" < "$prompt_file"` at pi-as-claude.sh:148), skill tree's gone (no `assets/pi/`, no stray refs). Tests pass (71 + 20). Also noticed you handled the stderr signal-token thing from last round, it's neutralized to `<<< RALPHEX:` now.

two small non-blockers, your call:

1. idle keepalive: a `text_delta` with no newline buffers but emits nothing (pi-as-claude.sh:187-190), so a long newline-free stream won't reset idle_timeout. Only bites if idle_timeout is set and the model streams for minutes without a newline, so low odds, but the comment claims keepalives cover idle. Emitting an empty keepalive on a buffered-but-unflushed delta would close it.

2. `PI_PROVIDER=anthropic`: the README says export the matching key before running ralphex, but ralphex strips `ANTHROPIC_API_KEY` unless `--preserve-anthropic-api-key`, so pi wouldn't see it for the anthropic provider. A one-line caveat in the README would cover it.

lgtm otherwise.

PR #382, opened 2026-06-10
feat: display user-set run parameters (models, executor) in web dashboard

### LINKED ISSUE #379: web ui: display configured run parameters (models, executor) in dashboard header
## Current behavior

The web dashboard header (`pkg/web/templates/base.html`) shows three pieces of run context: **Project**, **Plan**, and **Branch**.

## Proposed enhancement

It would be useful to also surface other user-defined parameters of the run, in particular the per-phase model configuration:

- `task_model` / `--task-model` — model[:effort] used for task execution
- `review_model` / `--review-model` — model[:effort] used for review phases (falls back to task model)
- `plan_model` / `--plan-model` — when running plan creation
- executor (claude vs codex), since `--codex` changes how the whole pipeline runs

When several sessions run in parallel (worktree mode) or when comparing runs after the fact, it's currently impossible to tell from the dashboard which model/effort a session was using — you have to remember what flags were passed on the CLI. Having this visible next to Plan/Branch (or in a small expandable details row) would make sessions self-describing.

Only parameters explicitly set by the user need to be shown; defaults could be omitted to keep the header compact.

Image

vlondon:
## What

Closes #379

Adds user-set run parameters (executor, plan/task/review model specs) to the web dashboard header, next to Project / Plan / Branch.

## How

Per the implementation guidance in the issue discussion, the values flow through the progress-file header so both display paths are covered:

- `pkg/progress/progress.go` — `writeHeader` writes optional `Executor:`, `Plan model:`, `Task model:`, `Review model:` lines next to `Plan:`/`Branch:`. Empty values are omitted entirely.
- `pkg/web/session_progress.go` — `ParseProgressHeader` parses the new lines into `SessionMetadata`, covering the watch / compare-after-the-fact view. Old progress files without the lines parse as before and simply show nothing.
- `pkg/web/server.go` — new `FormatRunParams` helper builds the display string (e.g. `codex · task gpt-5.5:high · review gpt-5.5:low`); exposed via the template for single-session `--serve` mode and as `runParams` in `/api/sessions` for multi-session mode.
- `base.html` / `app.js` / `styles.css` — new `Models` entry in the header info row, hidden when empty (`:empty`), updated on session switch and included in HTML export.
- `cmd/ralphex/main.go` — `runHeaderParams` resolves what gets recorded (CLI flag > config): only explicitly set values. The review model is recorded only when set directly (its implicit task_model fallback is not repeated); plan mode records the effective plan spec since `plan_model` falls back to `task_model` by design. `executor: codex` is recorded whenever the codex executor is active.

## Testing

- `go test -race ./...` and `golangci-lint run` clean; `GOOS=windows` cross-compile verified
- new table tests: header write (`TestNewLogger_HeaderRunParams`), header parse (incl. `Plan model:` not shadowing `Plan:`), `TestFormatRunParams`, `TestRunHeaderParams`
- existing index/sessions handler tests extended to assert the new field end-to-end
- manual verification: ran the watch-mode dashboard against fixture progress files (codex with both models, claude with task model only, old-format file) — header shows the params per session and hides the label entirely for old/default sessions

Screenshot 2026-06-10 at 09 54 47
Screenshot 2026-06-10 at 09 54 53
Screenshot 2026-06-10 at 09 55 02

### DIFF
--- a/cmd/ralphex/main.go
+++ b/cmd/ralphex/main.go
@@ -442,6 +442,7 @@ func setupProgressLogger(o opts, req executePlanRequest, branch string) (progres
 			Mode:           string(req.Mode),
 			Branch:         branch,
 			BranchOverride: req.BranchOverride,
+			Params:         runHeaderParams(o, req.Config, req.Mode),
 			NoColor:        o.NoColor,
 		}, req.Colors, holder)
 		if err != nil {
@@ -552,12 +553,14 @@ func executePlan(ctx context.Context, o opts, req executePlanRequest) error {
 	// wrap logger with broadcast logger if --serve is enabled
 	var runnerLog processor.Logger = plr.baseLog
 	if o.Serve {
+		params := runHeaderParams(o, req.Config, req.Mode)
 		dashboard := web.NewDashboard(web.DashboardConfig{
 			BaseLog:         plr.baseLog,
 			Port:            o.Port,
 			Host:            o.Host,
 			PlanFile:        req.PlanFile,
 			Branch:          branch,
+			RunParams:       web.FormatRunParams(params.Executor, params.PlanModel, params.TaskModel, params.ReviewModel),
 			WatchDirs:       o.Watch,
 			ConfigWatchDirs: req.Config.WatchDirs,
 			Colors:          req.Colors,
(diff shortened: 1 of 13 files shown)


umputun (approved):
lgtm. Both constraints from the issue are covered - params flow through the progress header so the watch / multi-session view gets them, not just the live dashboard - and tests/lint/race/windows are all clean.

one non-blocking note, no change needed here: on the restart-preserve path (re-running a failed or interrupted plan on the same progress file), `writeHeader` is skipped so the new run's params aren't re-recorded, and `ParseProgressHeader` reads only the first header block. So if you restart with different model flags, the dashboard would still show the original run's models/executor. It's display-only and consistent with how `Branch:`/`Mode:` already behave on restart, so fine to leave - just flagging for awareness.

PR #383, opened 2026-06-11
fix: plan panel marks finished tasks done and scrolls to the active task

### LINKED ISSUE #380: web ui: plan panel doesn't mark previous tasks as done, and doesn't scroll to the active task
## Description

In the web dashboard, the right-hand **Plan** side panel doesn't mark previous tasks as completed as the run progresses. When the run moves to a later task (e.g. Task 4), the panel highlights the current task as active (●), but Tasks 1–3 still show the pending circle (○) and their checkboxes remain unchecked — even though those tasks finished long ago (their sections in the main output area show durations like `7m 7s`, `1m 42s`, `4m 55s`).

After a **manual page refresh**, the previous tasks ARE correctly marked as completed in the plan panel — so the underlying state is right, the live SSE update path just doesn't apply it.

Possibly related: when a new task becomes active, the plan panel does **not** scroll to it. With a long plan, the active task is off-screen and the panel just sits wherever it was scrolled, so at a glance it looks like nothing is progressing (Task 4 is running but the panel still shows Task 1 at the top, unmarked).

## Observed behavior

- Run in progress, currently on Task 4 of 4
- Main output area correctly shows Tasks 1–3 as finished sections with durations
- Plan panel: Task 4 highlighted as active, but Tasks 1–3 still rendered as pending (○, unchecked checkboxes)
- Manually refreshing the page fixes it: completed tasks then show as done
- Plan panel viewport stays at the top (Task 1) instead of following the active task

## Expected behavior

- When task N+1 starts (or task N ends), task N should flip to done (✓) with its checkboxes checked in the plan panel — without requiring a page reload
- The panel should scroll the newly active task into view

## Code pointers

- `pkg/web/static/app.js`: `handleTaskEnd()` marks a task done only when a `task_end` event arrives; `clearActiveTasksExcept()` demotes a previously-active task back to **pending** unless all its checkboxes are already `.checked` — and checkbox state comes from the plan data fetched at page load, so a task that just finished falls back to ○ until the page is reloaded (which re-fetches the plan with `[x]` checkboxes)
- `pkg/web/static/app.js`: `handleTaskStart()` adds the `active` class but never calls `scrollIntoView` on the task element in the plan panel
- `pkg/web/broadcast_logger.go`: `task_end` is emitted on phase transition away from the task phase / on next task start — worth checking whether these events reach the client on the live SSE path

## Environment

- ralphex dashboard (web UI), live session via SSE

Image
Image

vlondon:
Fixes #380

## Root cause

Only `BroadcastLogger` (in-process runs) emitted `task_end` events. The progress-file parse paths never did:

- `loadProgressFileIntoSession` (replay when a page loads mid-run) emitted `task_start` per task via `buildPendingSectionEvents` but never a `task_end`, so on the client each new `task_start` left the previous task `.active` and `clearActiveTasksExcept` demoted it back to ○ pending — exactly the screenshots in #380.
- The Tailer's non-deferred path (`parseLine`, used when tailing live from an offset/end) emitted only `section` events — no task boundaries at all, so the active highlight never advanced live either.

A manual refresh "fixed" it because `/api/plan` re-reads the plan file with `[x]` checkboxes.

## Changes

**Backend**
- `buildPendingSectionEvents` now tracks the current task number and emits a `task_end` before the next `task_start`, or when a section moves past the task phase (review/codex/eval) — mirroring `BroadcastLogger` semantics (`onPhaseChanged` / `PrintSection`). A generic section that stays in the task phase does not end the task.
- `Tailer.parseLine` routes section lines through the same builder, so the live tail path gets `task_start`/`task_end` too.
- The active task number is carried across tailer restarts (`Tailer.CurrentTask()` / `TailerConfig.InitialTask` / `Session.lastTask`), following the existing `lastPhase`/`pendingSection` carryover pattern, so a reactivated tailer still ends the task that was active when the previous tailer stopped.

**Frontend (`app.js`)**
- `clearActiveTasksExcept`: a lower-numbered previously-active task is marked done (✓ with checked boxes) instead of being demoted to pending — tasks run sequentially, so losing active status to a later task means it finished. Higher-numbered tasks (plan edited mid-run / re-running an earlier task) keep the existing pending fallback.
- `updatePlanTaskStatus`: the newly active task is scrolled into view in the plan panel (`scrollIntoView`, same style as the existing section scroll).

## Testing

- New unit tests: `parseLine`/`parseLineDeferred` task boundary sequences (end-before-start, end-on-phase-transition, generic-section-keeps-task-active, `InitialTask` seeding), loader boundary sequence (`task_start:1, task_end:1, task_start:2, task_end:2`), and `lastTask` recording when a file ends mid-task.
- `go test -race ./...` green; `golangci-lint run` — 0 issues; `gofmt -l` clean.
- Manually verified end-to-end in watch mode (`ralphex --serve --watch <dir>`): simulated an 8-task run by appending task iteration sections to a progress file while connected. The SSE stream now carries a clean `task_start N → task_end N` pair for every task (verified with a raw `/events` capture; previously no `task_end` ever appeared on this path), and in the browser each finished task flips to ✓ with checked boxes live, the panel scrolls to the newly active task, and the last task flips to ✓ when the review phase starts — all without a page refresh. Reactivation was exercised too (appends to a session already marked completed), confirming the `lastTask` carryover.
- Not run: the playwright e2e suite (existing plan-panel assertions don't pin status values, so they're unaffected).

Screen Recording 2026-06-11 at 01 45 30


umputun (approved):
lgtm, the fix is correct and the tests cover it well.

couple of edge cases worth knowing about, both pre-existing (broadcast_logger does the same thing on in-process runs, so not introduced here), your call whether to handle now or in a follow-up:

1. the `task iteration N` section number is `NextPlanTaskPosition()`, not a monotonic counter, and `buildPendingSectionEvents` emits `task_end` for the current task before any next `task_start` regardless of whether N advanced. So a timeout/FAILED retry re-emits the same N and the panel flips that task to done then back to active. And if a later task gets unchecked mid-run so `NextPlanTaskPosition` goes backwards, the higher task gets marked done. `broadcast_logger.go:81-85` has the identical logic.

2. in `--tasks-only` the final task stays active. The file ends right after the last task with no following section to trigger its `task_end`, and the loader records `lastTask` at EOF without closing it (`session_progress.go:163`). Same gap in-process, `runTasksOnly` never transitions off `PhaseTask` so `onPhaseChanged` never fires the last `task_end` either. Emitting a `task_end` on the COMPLETED signal / Completed footer would close both paths at once.

two minor things:

- `lastTask` has `setLastTask` but no `getLastTask`, so the new tests read the field directly under a manual `RLock`. `lastOffset`/`lastPhase` both have getters, a `getLastTask()` would match the convention.
- `app.js:764` `scrollIntoView` uses `block:'start'`, which walks all scrollable ancestors and can nudge the whole page if the plan panel is off-screen. `block:'nearest'` would confine it to the panel.

none block the merge.

umputun:
following up on copilot's two notes, both are valid and worth folding in before merge:

- `tail.go:525` - agreed, `task_end` should carry `PhaseTask`, not the incoming section's phase. BroadcastLogger always tags it with the task phase (`onPhaseChanged` uses `old`, the task→task path is still in task phase), so hardcoding `status.PhaseTask` here matches. Doesn't affect the plan panel since `handleTaskEnd` only reads `task_num`, but a `task_end` tagged `PhaseReview` can bucket under the wrong section in the main output stream.
- `tail.go:514` - the named return `newCurrentTask` is never assigned and the body returns `currentTask` explicitly, so drop the name to keep the signature honest.

both quick. Ties into the parity point from my review above.

umputun (approved):
round-2 lgtm, all the requested changes folded in cleanly - task_end tagged PhaseTask, the forward-progress guard on the retry/backwards cases, tasks-only close on COMPLETED, getLastTask accessor, scrollIntoView nearest. Tests cover the new branches well.

one cosmetic edge it adds, non-blocking: a false completion (model emits ALL_TASKS_DONE while the plan still has [ ] items so the runner keeps going per task.go) briefly flips the task done then back to active on the next iteration. Self-corrects, fine to leave as-is. Merging.

PR #384, opened 2026-06-11
feat: add new Anthropic model Fable to supported agent models

Alex-Kopylov:
## What

Anthropic released a new model class — **Fable** (`claude-fable-5`), a tier above Opus.
Agent frontmatter validation accepted only `haiku`/`sonnet`/`opus`, so `model: fable`
(or a full ID like `claude-fable-5`) in an agent file was silently dropped with a warning.

## Changes

- `pkg/config/frontmatter.go` — add `fable` to `validModels`; update `Validate` warning
  text and `normalizeModel` doc comment
- `pkg/config/frontmatter_test.go` — new cases: `fable` keyword validation and
  `claude-fable-5` full-ID normalization; updated warning expectations
- `pkg/config/defaults/config`, `pkg/config/values.go`, `pkg/executor/executor.go`
  — extend model enumerations in comments
- `README.md`, `CLAUDE.md`, `llms.txt` — document `fable` as an accepted agent model
  keyword; update examples


umputun (approved):
lgtm, clean. `normalizeModel` picks up `claude-fable-5` for free since it scans the `validModels` keys, and that map is the only validation gate so coverage is complete. Tests cover both the keyword and full-ID paths.

one thing to verify on your end: this lets `model: fable` flow through to the Task-tool spawn, so worth confirming your claude install accepts `fable` as a `--model` shorthand the same way it does opus/sonnet/haiku. Not a blocker.

PR #400, opened 2026-07-03
fix: bound force-exit cleanup so a stuck worktree removal cannot prevent exit

paskal:
## Problem

The interrupt watcher's 5s force-exit path (`startInterruptWatcher`) calls the worktree cleanup directly before `os.Exit(1)`. That cleanup shares a `sync.Once` with the graceful shutdown's deferred cleanup in `runWithWorktree`, and `Once.Do` blocks until the first invocation completes. When the deferred cleanup is already in flight and stuck — e.g. `git worktree remove` hanging on a child process still holding the tree — the watcher prints `force exit` and then blocks forever inside `Once.Do`; the promised force exit never happens. This is exactly the scenario the 5s escape hatch exists for.

## Fix

Run the cleanup on the force-exit path through `runCleanupBounded`: execute it in a goroutine and wait at most 2s before proceeding to `os.Exit(1)` regardless (worst-case ~7s after cancel, noted in the function contract). A fast cleanup still completes normally; a stuck one gets a `cleanup did not finish in time, exiting anyway` notice instead of hanging the process.

Covered by a table-driven test for the nil / fast / stuck cleanup cases.

### DIFF
--- a/cmd/ralphex/main.go
+++ b/cmd/ralphex/main.go
@@ -1463,7 +1463,9 @@ func isResetOnly(o opts) bool {
 
 // startInterruptWatcher prints immediate feedback when context is canceled.
 // if graceful shutdown doesn't complete within 5 seconds, force exits.
-// cleanup, if not nil, is called only on the force-exit (5s timeout) path before os.Exit.
+// cleanup, if not nil, is called only on the force-exit (5s timeout) path before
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm - bounded-wait correctly breaks the Once.Do deadlock on the force-exit path, and the stuck-cleanup goroutine is harmless before os.Exit. clean test. thx

PR #406, opened 2026-07-13
fix: instruct default prompts to run long test commands in the foreground

### LINKED ISSUE #398: Bug: ralphex fails to wait for long test runs due to new CC behaviour
Ralphex consistently prematurely ending a phase failing to wait for a new CCs scheduled wake-up behaviour for long script runs.
Here is what output step looks like for my finalize step:
```
--- finalize step ---
[26-07-02 08:12:35] I'll start by understanding the current state of the branch and recent work, then run the test suite.
[26-07-02 08:14:16] Kicked off the full test suite (`./scripts/run-all-tests.sh`) in the background — it's currently running unit tests, then will move to integration, API, and
[26-07-02 08:14:16] ...
[26-07-02 08:14:16] ...
[26-07-02 08:14:26] ...
[26-07-02 08:14:26] ...
[26-07-02 08:14:44] ...
[26-07-02 08:14:47] ...
[26-07-02 08:14:58] Integration tests are still running with no output yet (jest buffers results until suites complete). Waiting for the scheduled check-in.
[26-07-02 08:15:10] Integration tests are progressing (Kafka topics being created for test runs). I'll hold off on further checks until the scheduled wakeup.
[26-07-02 08:15:17] Tasks are already up to date (#1 in progress, #2 done, #3-5 pending). Waiting for the scheduled wakeup now.
[26-07-02 08:15:27] Jest is actively running the integration tests (CPU active, ~70 node processes for parallel workers + backend services). This is expected to take several minutes given
[26-07-02 08:15:27] the size of the int test suite. Waiting for the scheduled wakeup.
[26-07-02 08:15:33] I'll now stop polling and wait for the scheduled wakeup at 08:26 or the background task completion notification, whichever comes first.
[26-07-02 08:15:38] I've scheduled a check-in for 08:26 and will report back once the test suite finishes or fails. No further action needed right now.
[26-07-02 08:15:41] Tasks remain accurate — no updates needed. Waiting for the scheduled wakeup at 08:26.
[26-07-02 08:15:45] Waiting for the test suite to progress further. I'll check back at the scheduled time.
[26-07-02 08:15:47] I'll stop checking now and wait for the scheduled wakeup or completion notification.
[26-07-02 08:15:52] Standing by for the scheduled wakeup at 08:26 — no further action needed until then.
[26-07-02 08:15:55] Ending this turn here — I'll resume automatically at the scheduled wakeup (08:26) or sooner if the background test run finishes first.
[26-07-02 08:16:01] finalize step completed
[26-07-02 08:16:01] all phases completed successfully
moved plan to docs/plans/completed/....md
```
It looks like it needs more heuristics on output parsing to correctly handle this.
Or probably better prompting...

mvanhorn:
## Summary

Wires the foreground-execution guidance from #398 into the four default prompt files at the exact points where they run tests / long commands:

- `pkg/config/defaults/prompts/finalize.txt` - Step 5 (Verify the branch is ready)
- `pkg/config/defaults/prompts/task.txt` - STEP 2 - VALIDATE
- `pkg/config/defaults/prompts/review_first.txt` - Step 3.3 (Fix All Confirmed Issues)
- `pkg/config/defaults/prompts/review_second.txt` - Path B (Issues found AND fixed)

Prompt text only, no Go changes. The embedded FS picks up the edited files at build time, and no markdown formatting is introduced (each file's OUTPUT FORMAT rule). `go build ./...` and `go test ./pkg/config/...` pass.

## Why this matters

ralphex drives the coding-agent CLI in headless, one-shot `-p` mode. Newer CLI behavior can push a long-running command (a full test suite) into the background, end the turn, and expect a scheduled wakeup to resume and collect results. Under one-shot `-p` there is no resume: the process exits, the backgrounded tests die with it, and ralphex sees the turn end and marks the step complete. The reporter's `finalize` log in #398 shows exactly this - the suite is started in the background, "waiting for the scheduled wakeup" lines follow, and "all phases completed successfully" is reported without the tests ever finishing.

In https://github.com/umputun/ralphex/issues/398#issuecomment-4950080427 you confirmed the mechanism and identified the practical fix as prompt-side - instruct the agent to run long verification commands in the foreground and block until they finish in the same turn - and named the exact files (`finalize.txt` step 5, `task.txt`, and the review prompts). This uses wording close to what you proposed.

You framed folding this into the defaults as conditional on the reporter confirming the fix holds. This is the exact guidance you proposed, wired into the four files you named - feel free to hold it until that confirmation if you'd prefer.

Closes #398

AI was used for assistance.

### DIFF
--- a/pkg/config/defaults/prompts/finalize.txt
+++ b/pkg/config/defaults/prompts/finalize.txt
@@ -32,6 +32,7 @@ Steps:
 5. Verify the branch is ready:
    - Run tests using the project's test command (check CLAUDE.md or plan file for the correct command)
    - Run linter if applicable
+   - Run tests and any long-running commands in the foreground and wait for them to finish in this turn. Do not run them in the background or end your turn expecting to resume at a scheduled wakeup - this session runs claude headless (claude -p) and does not resume.
 
 Report what was done. This step is best-effort - if rebase fails, explain why and the branch remains as-is.
 
--- a/pkg/config/defaults/prompts/finalize.txt
+++ b/pkg/config/defaults/prompts/finalize.txt
@@ -32,7 +32,7 @@ Steps:
 5. Verify the branch is ready:
    - Run tests using the project's test command (check CLAUDE.md or plan file for the correct command)
    - Run linter if applicable
-   - Run tests and any long-running commands in the foreground and wait for them to finish in this turn. Do not run them in the background or end your turn expecting to resume at a scheduled wakeup - this session runs claude headless (claude -p) and does not resume.
+   - Run tests and any long-running commands in the foreground and wait for them to finish in this turn. Do not run them in the background or end your turn expecting to resume at a scheduled wakeup - this session runs headless one-shot and does not resume.
 
 Report what was done. This step is best-effort - if rebase fails, explain why and the branch remains as-is.
 
--- a/pkg/config/defaults/prompts/review_first.txt
+++ b/pkg/config/defaults/prompts/review_first.txt
@@ -64,7 +64,7 @@ Do NOT reject issues just because they existed before this branch - fix them any
 
 ### 3.3 Fix All Confirmed Issues
 1. Fix all CONFIRMED issues (all types: bugs, tests, smells, docs, etc.)
-2. Run tests and linter to verify fixes - ALL tests must pass, ALL linter issues resolved. Run tests and any long-running commands in the foreground and wait for them to finish in this turn. Do not run them in the background or end your turn expecting to resume at a scheduled wakeup - this session runs claude headless (claude -p) and does not resume.
+2. Run tests and linter to verify fixes - ALL tests must pass, ALL linter issues resolved. Run tests and any long-running commands in the foreground and wait for them to finish in this turn. Do not run them in the background or end your turn expecting to resume at a scheduled wakeup - this session runs headless one-shot and does not resume.
 3. Commit fixes: `git commit -m "fix: address code review findings"`
 
 ## Step 4: Signal Completion
--- a/pkg/config/defaults/prompts/review_first.txt
+++ b/pkg/config/defaults/prompts/review_first.txt
@@ -64,7 +64,7 @@ Do NOT reject issues just because they existed before this branch - fix them any
 
 ### 3.3 Fix All Confirmed Issues
 1. Fix all CONFIRMED issues (all types: bugs, tests, smells, docs, etc.)
-2. Run tests and linter to verify fixes - ALL tests must pass, ALL linter issues resolved
+2. Run tests and linter to verify fixes - ALL tests must pass, ALL linter issues resolved. Run tests and any long-running commands in the foreground and wait for them to finish in this turn. Do not run them in the background or end your turn expecting to resume at a scheduled wakeup - this session runs claude headless (claude -p) and does not resume.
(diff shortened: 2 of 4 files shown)


umputun (commented):
before this goes into the shipped defaults I want proof it actually fixes the behavior, not just that it reads right. The mechanism is still my hypothesis, I said on #398 I hadn't hit it myself, and nobody has confirmed the prompt line resolves it yet (the reporter hasn't come back). A prompt instruction doesn't necessarily override CC's auto-backgrounding, that's the part that needs proving.

what would convince me: a before/after on a real run. One where a finalize or task step backgrounded the suite and reported success early, then the same run with this line where the tests actually block to completion in the same turn. Without that it's a plausible idea, not a verified fix, and I don't want the defaults shipping on a plausible idea.

second, the `(claude -p)` wording is too specific. These four prompts aren't claude-only, they also drive the codex executor (`--codex`) and the alternative-provider wrappers, so hardcoding "this session runs claude headless (claude -p)" is wrong for those runs. Keep the rationale agent-neutral like the original snippet, e.g. "this session runs headless one-shot and does not resume", so it holds no matter what's driving.

I'll fold it into the defaults once #398 is confirmed fixed. Until then it works fine as user-side prompt guidance, anyone hitting this can drop it into `~/.config/ralphex/prompts` or a project `.ralphex/prompts` today.

mvanhorn:
Thanks @umputun, both points are fair. On the wording: agreed — I'll drop the claude-specific `(claude -p)` phrasing and keep it agent-neutral (e.g. "this session runs headless one-shot and does not resume") so it holds for the `--codex` and alt-provider executors too. On proof: also fair — I'll put together a before/after on a real run (a finalize/task step backgrounding the suite and reporting success early, then the same run with this line where the tests block to completion in the same turn). Until I can show that, I agree it stays user-side prompt guidance rather than a shipped default. I'll follow up here with the evidence.

mvanhorn:
Wording change pushed in 382ff6e — the four prompts now say "this session runs headless one-shot and does not resume," agent-neutral so it holds for the `--codex` and alt-provider runs too. Before/after proof still to come.

mvanhorn:
Followed up on the proof you asked for, and the honest result is I can't produce it — because I can't reproduce the failure on current tooling. I ran a controlled before/after: a task step told to run a long test command (a ~200s streaming "integration suite" that only writes its pass marker at the very end), measuring whether the agent emits its completion token *before* the suite actually finishes.

Across four runs it never did:
- `claude -p` (CC 2.1.210), 130s test, no instruction → ran foreground, waited
- `claude -p`, same test, **with** the line → waited
- `claude -p`, 200s streaming suite, no instruction → waited
- `codex exec`, 200s streaming suite, no instruction → waited

Every time, the agent ran the suite in the foreground and only reported complete after it truly finished — with and without the line, on both executors. So I can't show a "before" where the suite is backgrounded and success is reported early, which means I can't show the line preventing it. On current claude/codex the auto-background-then-report-early pattern doesn't seem to fire for a foreground-invoked test command, so the line is harmless belt-and-suspenders here rather than a demonstrable fix — which matches your read.

If you have the repro recipe from the original report (the command shape / agent / version that backgrounded), I'll run exactly that. Otherwise I'd keep it as opt-in prompt guidance rather than a shipped default, since I can't back the "it fixes X" claim with evidence.

PR #414, opened 2026-07-19
fix: align default review agents with the review-loop contract

alekb:
## What

Tightens the five default review agent prompts (`pkg/config/defaults/agents/`) and the review-context lead-in so the spawned reviewers actually honor the contract the review loop grades them against.

## Why

I audited the five default agent bodies against how `expandAgentReferences`/`formatAgentExpansion` compose them and what `review_first.txt`/`review_second.txt` expect back. The bodies describe *what* to review well, but omit the operational contract the loop depends on, and each omission converts directly into review-iteration churn:

- **No diff scoping.** Checklists read as repo-wide audit mandates ("all user inputs validated", "functions or branches without test coverage", the whole over-engineering taxonomy). Combined with the parent's "Do NOT reject issues just because they existed before this branch - fix them anyway", agents can surface pre-existing backlog every iteration, making the zero-findings `REVIEW_DONE` exit hard to reach on mature repos.
- **No clean-case output.** Every body ends with "report problems only - no positive observations" and never says what to emit when there is nothing to report - an agent with zero findings literally cannot comply, and an empty reply is indistinguishable from a truncated run. The `NO ISSUES FOUND` sentinel already exists for external reviewers (`custom_review.txt`, `codex_review.txt`); the spawned agents just never adopted it.
- **Nothing says read-only.** The reviewers run as full-tooled `general-purpose` subagents, five in parallel, while the parent owns fix-and-commit (Step 3.3). `documentation.txt` went further and *instructed* edits ("Mark completed items as done", "Update plan status") - mid-review plan mutations by a parallel reviewer, bypassing the verification gate, and duplicating what `task.txt` Step 3 already does before review starts.
- **The goal reviewer never gets the goal.** `implementation.txt` verifies "the stated goal/requirement", but nothing in the composed prompt states it - the agent must reverse-engineer requirements from commit messages. Variable expansion inside agent bodies is already supported (`replaceBaseVariables` in `expandAgentReferences`); the defaults just never used it.
- **No severity signal for the second pass.** `review_second.txt` filters to critical/major, but that instruction sits outside the quoted agent prompt, and neither `quality` nor `implementation` reports a severity - the parent re-derives it from prose.
- **Overlap between agents.** `quality.txt`'s Simplicity Assessment restates the dedicated simplification agent (which runs in the same parallel batch), and its findings are almost never critical/major in the second pass; `implementation.txt`'s edge-case item duplicated quality's; a third of `testing.txt` restated its own other sections. Cross-agent duplicates are dedup work every iteration.

## Changes

**`pkg/processor/prompts.go`** - `reviewContextInstruction` now carries the shared reviewer contract in the lead-in (so user custom agents inherit it too): diff scope, read-only, evidence bar, `NO ISSUES FOUND` sentinel.

**`implementation.txt`** - gets `{{GOAL}}`/`{{PLAN_FILE}}` (graceful fallbacks exist in plan-less mode); edge-case item rescoped to requirement-level scenarios (generic boundary bugs are quality's domain); report format aligned with quality's (exact file:line first) + `Severity`; closing line no longer accidentally narrows the mandate to one of the six responsibilities.

**`quality.txt`** - Simplicity Assessment replaced with a one-line exclusion pointing at the simplification agent (also matches the README's documented scope for this agent: "bugs, security issues, race conditions"); internal duplicates removed (concurrency owned by item 5, validation/sanitization by Security 1); + `Severity`.

**`testing.txt`** - scoped to changed code; Test Quality collapsed to its two unique items; location field now covers missing-test findings (the untested *source* location); skipped/disabled tests added; explicitly allowed to run the suite read-only; de-Go-ified the `_` idiom; + `Severity` and a focus line.

**`simplification.txt`** - scope + plan-awareness ("complexity explicitly required by the plan is not a finding"); 'unused / no callers / never triggers' claims require a cited project-wide search (they're whole-repo absence claims an agent pointed only at the diff will otherwise guess); skips generated/vendored code (this repo's own moq mocks pattern-match "pass-through wrapper" by construction); producer-side-interface bullet qualified to structurally-typed languages; large-effort findings recommended as follow-up plans rather than mid-review refactors.

**`documentation.txt`** - Plan Files bullets converted from edit imperatives to report items anchored on `{{PLAN_FILE}}`; reads current README/CLAUDE.md before reporting a gap; also catches docs the change made *stale* (not just missing ones); findings get a `Trigger: file:line` field so the parent's file:line-based verification and dedup can process them.

**`prompts_test.go`** - new subtest for the contract lines; the all-five-agents expansion test now compares against the body after `replaceBaseVariables`, since the defaults now legitimately use template variables.

## Testing

- `go test ./...` - all packages pass
- `golangci-lint run --max-issues-per-linter=0 --max-same-issues=0` - 0 issues
- Verified composed prompts still satisfy the existing content assertions (e.g. second-pass prompt contains no "test coverage" text, agent bodies carry the lead-in in both claude and codex shapes)


umputun (commented):
solid PR, no blocking issues. Build/tests/lint/race pass and it composes correctly for both the claude and codex agent shapes. Answers to your two questions, then two small optional notes.

**contract in Go vs the txt files:** keep it in `reviewContextInstruction`. It matches what `codexReviewGuidance`/`codexTaskGuidance` already do - Go-level injection so the directive still reaches users who replaced the embedded review prompts. Same reasoning applies to the reviewer contract, so no need to move it into the txt files; the non-overridability is the intended trade-off.

**pre-existing-issues tension you flagged:** it holds up. The diff-scoping doesn't suppress failing tests - `testing.txt` still reports suite failures, and the parent re-runs tests+lint after any fix, so a real failure still surfaces. As for whether the churn shows up in real runs, the scenarios are plausible from the prompts, but the sentinel and read-only fixes are worth it on their own regardless.

two optional notes:

- **scope-creep detection lost its owner.** `quality.txt`'s old "changes solve only the stated problem" bullet went away with the Simplicity Assessment section and didn't land elsewhere - `implementation.txt` checks whether the goal was met (did it do enough), not whether unrelated behavior was added (did it do too much), and `simplification.txt` is over-engineering only. Looks like collateral from dropping that section. A one-line bullet in `implementation.txt` about unrelated/unintended changes beyond the requirement would close it.
- **duplicate closing line.** After aligning each body's last line to `Report findings only - no positive observations.`, that line now appears twice in the composed prompt - once in the body, once appended by `formatAgentExpansion`. Drop it from the five bodies or from the wrapper, not both.

`testing.txt` doesn't lose the test-independence check, in case that looked risky - the removed line duplicated the untouched `## Test Independence` section.

alekb:
Thanks for the review. Both notes addressed in db32f75:

- scope-creep check restored — you're right, that was collateral from dropping the Simplicity Assessment section. It's now responsibility 7 in `implementation.txt` ("unrelated or unintended changes beyond the stated requirement"), which is arguably a better home than quality anyway since goal-vs-diff is that agent's whole frame.
- closing line deduped — removed from the five bodies, kept the `formatAgentExpansion` suffix so it also covers custom agents.

Good to hear on keeping the contract in `reviewContextInstruction` — the `codexReviewGuidance` parallel makes sense.

Tests and lint still pass after the master merge + these changes.

umputun:
both notes resolved, verified. Responsibility 7 in `implementation.txt` is the right home for the scope-creep check - goal-vs-diff is that agent's frame anyway. Closing line now appears once (bodies stripped, `formatAgentExpansion` suffix kept, so custom agents still get it). Pulled the head and ran it: full suite + lint green. Nothing new on my end - good to merge.

PR #420, opened 2026-07-26
Update README.md

vishfulthinking:
Updated model from gpt-5.5 to gpt-5.6


umputun (commented):
not merging this as-is, it makes the README wrong rather than right.

`pkg/config/defaults/config` still ships `codex_model = gpt-5.5`, so the two README spots match what ralphex actually defaults to. Editing only the docs makes them disagree.

the default lives in the embedded config, so bumping it is a code change: the value and the three comment lines in `pkg/config/defaults/config`, the expectations in `pkg/config/config_test.go` and `pkg/config/values_test.go`, `llms.txt`, then the two README spots.

also `gpt-5.6` in the PR body vs `gpt-5.6-sol` in the diff, worth settling which one it is.

Issue #355, opened 2026-05-22
RFC: orthogonal named executors, roles, pools, and config deprecation path


alex-mextner:
## Context

PR #350 added first-class `executor = codex` support and moved the runner toward role-based execution internally. That is a good foundation for making execution provider-neutral, but the public config is still a mix of provider-specific global keys:

```ini
claude_command = claude
claude_args = ...
task_model = ...
review_model = ...

codex_command = codex
codex_model = gpt-5.5
codex_reasoning_effort = xhigh
codex_sandbox = read-only

external_review_tool = codex
custom_review_script = ...
```

These keys work, but they do not scale well once users want to combine Claude, Codex, OpenCode/DeepSeek, Kimi, GLM, wrappers, or other command-backed tools across task and review roles.

## Intent

Introduce an orthogonal named executor model:

- executors are user-named;
- only semantic roles are `task` and `review`;
- `finalize` uses the `review` executor by default;
- no public `external` role;
- no public `custom` role;
- no provider enum like `claude|codex|script` in the public executor model;
- built-in `claude` and `codex` executors are always available;
- user-defined `[executor.<name>]` sections either overlay built-ins by name or create new command-backed executors.

Provider identity should not be part of the public model. The public model should describe what Ralphex needs to run: command, args, role, mode, concurrency, and I/O contract if needed.

## Proposed config shape

Built-ins exist implicitly:

```ini
# implicit built-ins
# executor.claude
# executor.codex
```

Users can tune built-ins by overlaying them:

```ini
[executor.claude]
command = claude
args = --dangerously-skip-permissions --output-format stream-json --verbose
task_model = sonnet
review_model = opus:xhigh

[executor.codex]
command = codex
model = gpt-5.5
reasoning_effort = xhigh
sandbox = danger-full-access
pass_claude_md = true
```

Users can define arbitrary executors:

```ini
[executor.deepseek]
role = review
mode = audit
command = opencode
args = run --model deepseek-reasoner ...

[executor.kimi]
role = task
mode = edit
command = opencode
args = run --model kimi-k2.6-turbo ...
```

Role selection:

```ini
task_executors = claude,kimi
task_parallel = 1

review_executors = codex,deepseek
review_parallel = 1
```

Defaults should preserve current behavior:

```ini
task_executors = claude
task_parallel = 1

review_executors = claude
review_parallel = 1
```

When `executor = codex` is used for backward compatibility, it can map to:

```ini
task_executors = codex
review_executors = codex
```

[...]

Issue #59, opened 2026-02-02
feat: add Gemini as alternative external reviewer


umputun:
**Summary**

Add Gemini CLI as an alternative to Codex for external code review phase. User can select which tool to use via config.

**Motivation**

From #57 discussion - some users prefer Gemini for code review. Current codex integration has hardcoded argument structure that doesn't work with Gemini CLI.

**Proposed Design**

**Config changes:**
```ini
# new selector (default: codex for backward compatibility)
review_tool = codex          # codex | gemini | none

# existing codex config (unchanged)
codex_enabled = true
codex_command = codex
codex_model = gpt-5.2-codex
codex_reasoning_effort = xhigh
codex_timeout_ms = 3600000
codex_sandbox = read-only

# new gemini config (parallel structure)
gemini_command = gemini
gemini_model = gemini-2.5-flash
gemini_timeout_ms = 3600000
gemini_sandbox = true
gemini_error_patterns = rate limit,quota
```

**Gemini CLI usage (confirmed working):**
```bash
gemini --sandbox --model gemini-2.5-flash --output-format stream-json "prompt"
```

**Key differences from Codex:**

| Aspect | Codex | Gemini |
|--------|-------|--------|
| Sandbox | `--sandbox read-only` (string) | `--sandbox` (boolean flag) |
| Model | `-c model=X` | `--model X` |
| Output | stderr (progress) + stdout (response) | stdout only (JSON lines) |
| Streaming | Custom filter (headers, bold text) | Parse JSON, extract `content` |

**Stream-JSON format:**
```json
{"type":"init","session_id":"...","model":"gemini-2.5-flash"}
{"type":"message","role":"user","content":"..."}
{"type":"message","role":"assistant","content":"chunk1","delta":true}
{"type":"message","role":"assistant","content":"chunk2","delta":true}
{"type":"result","status":"success","stats":{...}}
```

**Implementation tasks:**
1. Add `review_tool` config option with validation
2. Add `gemini_*` config fields (command, model, timeout, sandbox, error_patterns)
3. Create `GeminiExecutor` in `pkg/executor/gemini.go`
4. Update `Runner` to select executor based on `review_tool`
5. Make logging headers dynamic (`gemini external review` vs `codex external review`)
6. Update documentation

**Backward compatibility:**
- Default `review_tool = codex`
- `codex_enabled = false` still works (overrides selector)
- All existing codex config unchanged

umputun:
[...]

PR #429, opened 2026-08-10
feat: add orcarouter claude provider to docker wrapper

Marc-oss-hub:
## Summary

Adds [OrcaRouter](https://www.orcarouter.ai) as a Claude provider for the Docker wrapper. OrcaRouter is an OpenAI-compatible model routing gateway that exposes 150+ models from OpenAI, Anthropic, Google, DeepSeek and others behind a single endpoint and API key. It also provides gateway-level security controls for AI agents.

I'm an engineer on the OrcaRouter team.

## What this changes

- `scripts/ralphex-dk.sh`: adds `--claude-provider=orcarouter` / `RALPHEX_CLAUDE_PROVIDER=orcarouter` as a third provider mode. Translates `ORCAROUTER_API_KEY` to `ANTHROPIC_AUTH_TOKEN` via the docker inherit form (the key value never appears on the command line), auto-sets `ANTHROPIC_BASE_URL=https://api.orcarouter.ai`, and pins `anthropic/`-namespaced default model vars (bare Anthropic model ids fail on the gateway with `model_not_found`). Skips claude_home/keychain checks like the bedrock path.
- `scripts/ralphex-dk/ralphex_dk_test.py`: 13 unit tests covering provider selection, env translation, base-url/model pinning, `-E` overrides, and no-duplicate `-e` flags.
- `docs/orcarouter-setup.md`: setup guide mirroring `docs/bedrock-setup.md`.
- `README.md`, `llms.txt`, `CLAUDE.md`: usage notes and env var documentation.

## Why a separate provider rather than the OpenAI-compatible escape hatch

This mirrors how AWS Bedrock is wired in this repo — the wrapper already treats Claude providers as first-class `--claude-provider` modes, so OrcaRouter slots into the same flag, diagnostics, and env-passing path.

## How to use it

1. Get an API key at https://www.orcarouter.ai/console (keys start with `sk-orca-`).
2. Set `ORCAROUTER_API_KEY`.
3. Run:

   ```bash
   ralphex --claude-provider=orcarouter docs/plans/feature.md
   ```

   The wrapper auto-sets the gateway base URL and `anthropic/`-namespaced model pins; each can be overridden per-run with `-E`.

## Verification

- Unit tests: 13 `TestOrcaRouterProvider` tests pass (full wrapper suite matches the pre-change baseline: `failures=5, errors=28`, all pre-existing Windows-host environment issues unrelated to this change).
- Live API: all four auto-set model pins (`anthropic/claude-opus-5`, `anthropic/claude-sonnet-5`, `anthropic/claude-haiku-4.5`) return `ORCA-OK` from the Anthropic-compatible `/v1/messages` endpoint using the wrapper env form.
- Ran locally: wrapper `--dry-run --claude-provider=orcarouter` builds the docker command with the gateway base URL, model pins, and the auth token passed via the inherit form.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -129,6 +129,14 @@ Key functions in `scripts/ralphex-dk.sh`:
 - `export_aws_profile_credentials()` - exports credentials from AWS profile
 - `validate_bedrock_config()` - validates bedrock config, returns warnings
 
+### OrcaRouter Provider (Docker Wrapper Only)
+
+`scripts/ralphex-dk.sh` supports [OrcaRouter](https://www.orcarouter.ai) as a Claude provider (`--claude-provider orcarouter` / `RALPHEX_CLAUDE_PROVIDER`). See `docs/orcarouter-setup.md`.
+
+Key functions in `scripts/ralphex-dk.sh`:
+- `get_claude_provider()` - returns provider from CLI flag or env var
+- `build_orcarouter_env_args()` - builds docker -e flags for ORCAROUTER_ENV_VARS; translates `ORCAROUTER_API_KEY` to `ANTHROPIC_AUTH_TOKEN` via the inherit form, always sets `ANTHROPIC_BASE_URL=https://api.orcarouter.ai`, and pins `anthropic/`-namespaced default model vars (bare Anthropic model IDs fail on the gateway with `model_not_found`)
+
 ### Docker Socket Support (Docker Wrapper Only)
 
 `--docker` flag (or `RALPHEX_DOCKER_SOCKET`) mounts the host Docker socket for testcontainers. Socket path from `DOCKER_HOST` (unix://) or `/var/run/docker.sock`; GID auto-detected and passed via `DOCKER_GID`. Missing socket = fail-fast error.
--- a/README.md
+++ b/README.md
@@ -497,6 +497,30 @@ ralphex docs/plans/feature.md
 
 See [Bedrock setup documentation](docs/bedrock-setup.md) for detailed IAM policies and setup instructions.
 
+**OrcaRouter support:**
+
+When `--claude-provider=orcarouter` or `RALPHEX_CLAUDE_PROVIDER=orcarouter` is set:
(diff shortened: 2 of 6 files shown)


umputun:
not merging this as a vendor-named provider mode.

the wrapper already does what the branch adds. `-E`/`RALPHEX_EXTRA_ENV` passes arbitrary env into the container, and `claudeChildEnv` strips only `ANTHROPIC_API_KEY` and `CLAUDECODE`, so this works on master today for anyone with a local claude install:

```bash
export ORCAROUTER_API_KEY=sk-orca-...
export ANTHROPIC_AUTH_TOKEN="$ORCAROUTER_API_KEY"
ralphex -E ANTHROPIC_BASE_URL=https://api.orcarouter.ai \
        -E ANTHROPIC_AUTH_TOKEN \
        -E ANTHROPIC_DEFAULT_OPUS_MODEL=anthropic/claude-opus-5 \
        -E ANTHROPIC_DEFAULT_SONNET_MODEL=anthropic/claude-sonnet-5 \
        -E ANTHROPIC_DEFAULT_HAIKU_MODEL=anthropic/claude-haiku-4.5 \
        -E ANTHROPIC_SMALL_FAST_MODEL=anthropic/claude-haiku-4.5 \
        docs/plans/feature.md
```

the `-e` set that produces is the same one `--claude-provider=orcarouter` produces. So the branch buys a shorter command line, and in exchange the repo takes on a hardcoded endpoint plus four model ids that only OrcaRouter can keep current. When one of those changes I get the bug report and no way to check the fix.

same call as #59 and #355: third party tools go through the generic seams (`claude_command`/`claude_args`, `-E`, `external_review_tool = custom`), not a named mode. `--claude-provider` covers Claude Code's own auth backends, which is why bedrock sits there and a routing gateway doesn't.

two things on my side, unrelated to this branch: the wrapper won't start without `~/.claude` (`scripts/ralphex-dk.sh:1121`), so the command above needs a local login it shouldn't need, and the `ANTHROPIC_BASE_URL` route isn't documented anywhere. Both are mine to fix.

PR #433, opened 2026-08-18
fix(pi-as-claude): pass max thinking level through to pi

asalimonov:
pi added a native `max` thinking level (https://github.com/earendil-works/pi/issues/6097#issuecomment-4946495274), but the pi-as-claude wrapper still downgraded ralphex's `max` effort to `xhigh` with a stderr note. this removes the special case: all effort levels pass through to pi's `--thinking` verbatim and pi validates them. docs and mock-pi tests updated accordingly.

verified with ralphex 1.6.0 and pi 0.84.1: `--task-model="k3:max"` reaches pi as `--thinking max`; wrapper tests 70/70, shellcheck clean.

### DIFF
--- a/docs/custom-providers.md
+++ b/docs/custom-providers.md
@@ -381,8 +381,7 @@ ralphex appends `--model <m>` / `--effort <e>` per phase. The wrapper forwards `
 
 | ralphex effort | pi thinking |
 |---|---|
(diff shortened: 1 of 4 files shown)


umputun (approved):
lgtm

PR #454, opened 2026-09-01
fix: return repo-relative paths with forward slashes so the plan file excludes itself on Windows

### LINKED ISSUE #453: worktree mode: plan file cannot exclude itself on Windows (path separator mismatch in toRelative)
## Summary

On Windows, `--worktree` refuses to start with a message that contradicts itself: it reports uncommitted changes "other than the plan file" and then lists the plan file as the only entry.

```
error: create worktree: cannot create worktree: worktree has uncommitted changes other than the plan file

uncommitted files:
  docs/plans/my-plan.md
```

The working tree contains nothing else — `git status --porcelain -uall` returns exactly that one line.

## Environment

- ralphex v1.6.1 (built from source with `go install github.com/umputun/ralphex/cmd/ralphex@latest`; the releases carry no Windows asset)
- Go 1.27.0, windows/amd64
- git 2.55.0.windows.3
- Windows 11

## Steps to reproduce

1. On Windows, in a clean checkout on the default branch.
2. Write a plan to `docs/plans/my-plan.md` (the default `plans_dir`) and leave it uncommitted.
3. `ralphex --worktree docs/plans/my-plan.md`

**Expected:** the run starts; the plan file is copied into the worktree and committed there (`pkg/git/service.go`).
**Actual:** the run aborts with the message above, naming the plan file as the offending change.

Passing the path with backslashes does not help — `filepath.Clean` normalizes both spellings to backslashes.

## Cause

`toRelative` (`pkg/git/external.go:510`) returns the result of `filepath.Clean` / `filepath.Rel` unchanged, so on Windows `docs/plans/my-plan.md` becomes `docs\plans\my-plan.md`.

`hasChangesOtherThan` (`pkg/git/external.go:298`) compares that value against the paths from `git status --porcelain`, and git prints forward slashes on every platform:

```go
filePath := e.extractPathFromPorcelain(line)   // "docs/plans/my-plan.md"
if strings.EqualFold(filePath, rel) {          // rel == "docs\plans\my-plan.md"
    continue
}
```

`EqualFold("docs/plans/my-plan.md", "docs\plans\my-plan.md")` is false, so the plan file can never exclude itself once it lives in a subdirectory — and `plans_dir` defaults to `docs/plans`. On Windows, `--worktree` with an uncommitted plan file is therefore blocked in general.

The existing tests already show it. On Windows, unchanged `master` fails:

```
--- FAIL: TestExternalBackend_toRelative/returns_repo-relative_path_unchanged
        expected: "docs/plans/test.md"
        actual  : "docs\plans\test.md"
--- FAIL: TestExternalBackend_HasChangesOtherThan/returns_empty_when_only_target_file_is_untracked
--- FAIL: TestExternalBackend_HasChangesOtherThan/returns_dirty_file_when_other_file_is_untracked
--- FAIL: TestExternalBackend_HasChangesOtherThan/returns_dirty_file_when_tracked_file_is_modified
--- FAIL: TestExternalBackend_HasChangesOtherThan/excludes_plan_file_with_different_case
```

## Workaround

Commit the plan file before starting; ralphex then takes it from HEAD and the tree is clean. Verified.

## Suggested fix

Convert the return values of `toRelative` with `filepath.ToSlash`. Every other caller (`add`, `moveFile`, `commitFiles`, `fileHasChanges`) hands the result to git, which accepts forward slashes on all platforms, so the conversion is safe there as well.

I have a branch with the fix and the test adjustments and will open a PR against this issue.

dimi4ik:
Fixes #453.

## What

`toRelative` (`pkg/git/external.go`) returned the output of `filepath.Clean` / `filepath.Rel` unchanged, so on Windows a repo-relative path came back as `docs\plans\my-plan.md`. `hasChangesOtherThan` compares that against the paths from `git status --porcelain`, which are forward-slashed on every platform, so the `strings.EqualFold` check never matched and the plan file could not exclude itself once it lived in a subdirectory — `plans_dir` defaults to `docs/plans`, so `--worktree` with an uncommitted plan file was blocked on Windows in general.

Both return paths now go through `filepath.ToSlash`. The other callers (`add`, `moveFile`, `commitFiles`, `fileHasChanges`) pass the value straight to git, which accepts forward slashes everywhere, so the conversion is safe for them too.

The second commit fixes test expectations, not code: `TestService_MovePlanToCompleted` built its expected paths with `filepath.Join` and matched them against the output of `git show --name-status` and `git ls-tree`. Those strings come from git, so the expectation side was wrong — those two cases fail on Windows both before and after the fix, and without the change `go test ./pkg/git/` cannot go green there.

## How tested

`go test ./pkg/git/` on Windows 11, Go 1.27.0, git 2.55.0.

Before, on unchanged `master`, six subtests fail:

```
--- FAIL: TestExternalBackend_toRelative/returns_repo-relative_path_unchanged
        expected: "docs/plans/test.md"
        actual  : "docs\plans\test.md"
--- FAIL: TestExternalBackend_toRelative/rejects_absolute_path_outside_repo
--- FAIL: TestExternalBackend_HasChangesOtherThan/returns_empty_when_only_target_file_is_untracked
--- FAIL: TestExternalBackend_HasChangesOtherThan/returns_dirty_file_when_other_file_is_untracked
--- FAIL: TestExternalBackend_HasChangesOtherThan/returns_dirty_file_when_tracked_file_is_modified
--- FAIL: TestExternalBackend_HasChangesOtherThan/excludes_plan_file_with_different_case
```

After: `ok github.com/umputun/ralphex/pkg/git`, `golangci-lint run ./pkg/git/...` reports 0 issues, `gofmt` clean.

Test changes beyond the new case:

- `converts absolute path to relative` now expects forward slashes, matching the documented behaviour of `toRelative`.
- `rejects absolute path outside repo` used the literal `/tmp/outside/file.txt`, which `filepath.IsAbs` rejects on Windows — the case took the relative branch and never reached the check it names. It now uses `t.TempDir()`, absolute on every platform.
- New case `converts os-specific separators to forward slashes`, so a regression is caught on Linux too.

Note for anyone reproducing on Windows: run the suite with `core.autocrlf=false`, otherwise `TestService_CreateWorktreeForPlan` fails on a `\r\n` vs `\n` comparison of test fixture content. That is unrelated to this change.

### DIFF
--- a/pkg/git/external.go
+++ b/pkg/git/external.go
@@ -506,14 +506,17 @@ func (e *externalBackend) refExists(ref string) bool {
 	return cmd.Run() == nil
 }
 
-// toRelative converts a path to be relative to the repository root.
+// toRelative converts a path to be relative to the repository root, always with forward slashes.
+// git speaks forward slashes on every platform, both in its own output (status --porcelain) and in
+// the paths it accepts as arguments, while filepath.Clean and filepath.Rel yield backslashes on
+// windows - an unconverted result would never match the paths git reports back.
 func (e *externalBackend) toRelative(path string) (string, error) {
 	if !filepath.IsAbs(path) {
 		cleaned := filepath.Clean(path)
 		if strings.HasPrefix(cleaned, "..") {
 			return "", fmt.Errorf("path %q escapes repository root", path)
 		}
-		return cleaned, nil
+		return filepath.ToSlash(cleaned), nil
 	}
 
 	// resolve symlinks for consistent comparison (macOS /var -> /private/var)
@@ -531,7 +534,7 @@ func (e *externalBackend) toRelative(path string) (string, error) {
 	if strings.HasPrefix(rel, "..") {
 		return "", fmt.Errorf("path %q is outside repository root %q", path, e.path)
 	}
-	return rel, nil
+	return filepath.ToSlash(rel), nil
 }
 
 // addWorktree creates a git worktree at the given path.
--- a/pkg/git/external_test.go
+++ b/pkg/git/external_test.go
@@ -879,6 +879,12 @@ func TestExternalBackend_toRelative(t *testing.T) {
 		assert.Equal(t, "docs/plans/test.md", rel)
 	})
 
+	t.Run("converts os-specific separators to forward slashes", func(t *testing.T) {
+		rel, err := eb.toRelative(filepath.Join("docs", "plans", "test.md"))
+		require.NoError(t, err)
+		assert.Equal(t, "docs/plans/test.md", rel, "git reports forward slashes on every platform")
+	})
+
 	t.Run("rejects .. path", func(t *testing.T) {
 		_, err := eb.toRelative("../outside.txt")
 		require.Error(t, err)
@@ -889,11 +895,14 @@ func TestExternalBackend_toRelative(t *testing.T) {
 		absPath := filepath.Join(eb.path, "docs", "plans", "test.md")
 		rel, err := eb.toRelative(absPath)
 		require.NoError(t, err)
-		assert.Equal(t, filepath.Join("docs", "plans", "test.md"), rel)
+		assert.Equal(t, "docs/plans/test.md", rel)
 	})
 
 	t.Run("rejects absolute path outside repo", func(t *testing.T) {
-		_, err := eb.toRelative("/tmp/outside/file.txt")
+		// t.TempDir is absolute on every platform; a unix-style literal would fall into the
+		// relative branch on windows and never reach the check under test
+		outside := filepath.Join(t.TempDir(), "file.txt")
+		_, err := eb.toRelative(outside)
 		require.Error(t, err)
 		assert.Contains(t, err.Error(), "outside repository")
 	})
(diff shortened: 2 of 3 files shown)


umputun:
merged. This report made that possible: it named the cause, the failing tests, and the fix.

worth saying why this one is in, because Windows changes usually cannot be. I have no Windows machine, so I can neither reproduce a Windows report nor verify a Windows fix. This one is clear-cut: the cause is unambiguous, the change is small and self-contained, and `filepath.ToSlash` returns its argument unchanged when the separator is already `/`, so Linux and macOS cannot behave differently. The Windows runtime result rests on your run. The code makes the conversion and Unix no-op clear, but I cannot execute the Windows path.

`09fe05e` now documents that position in `README`, `llms.txt`, and `CLAUDE.md`: Windows is best effort, with no Windows release binary or maintainer testing and a narrow bar for Windows-only issues and PRs.

one extra scope note: on current master the same mismatch also blocked plain branch mode. `CreateBranchForPlan` calls `preparePlanBranch` with `worktreeMode=false`, where the plan was still treated as unrelated dirt.

PR #455, opened 2026-09-01
fix: match claude limit patterns by diagnostic provenance

### LINKED ISSUE #452: claude limit patterns match the agent quoting them, not just the CLI emitting them
## What happens

With `wait_on_limit` set, a **successful** claude run can put ralphex into a wait/retry loop that never converges.

The run that triggered it here was a review phase that had genuinely been killed by a session limit earlier in the evening. The retry ran fine — it executed the build and tests, produced a correct report, and exited zero. But its report explained what had gone wrong, and in doing so quoted the phrase:

```
[26-08-26 20:40:01]   - Three of the four lenses (bugs+impl, docs+tests,
[26-08-26 20:40:01] arch+quality) were cut off mid-investigation: "rate limited:
[26-08-26 20:40:01] rejected / You've hit your session limit · resets 8:30pm
[26-08-26 20:40:01] (Europe/Minsk)", then "degraded". None of them answered.
...
[26-08-26 20:40:01] rate limit detected: "You've hit your session limit" in claude output, waiting 2h0m0s before retry...
```

The last line is ralphex firing on the line four above it. The whole block is one agent's report of an earlier failure, written by a run that then exited zero.

## Why config can't fix it

This was my first assumption too, and I don't think it holds. The pattern and the false positive are the same string. Any pattern specific enough to match the real CLI message also matches an agent quoting that message in a successful report — and agents quote error text constantly, because that is what a good failure report looks like.

Narrowing the phrase doesn't help, and dropping it means real limits stop being detected. The missing information isn't vocabulary, it's **who said it**.

The 10-block `RecentText` window bounds how far back stale text survives, but it doesn't help here: the offending quote is in the run's *final* response.

## Why I think this is in scope

`llms.txt` already describes the same class of bug on the codex side as fixed:

> The codex defaults are tightened so that review findings that *talk about* rate limiting in a codebase do not trip a false positive.

That was solved by tightening vocabulary, which works when the phrases are generic enough to narrow. The claude phrases aren't — `You've hit your session limit` is exact CLI text with no narrower form.

I also read #281 before writing any of this, and I think the objection there was right: `non-empty stdout = complete output` isn't verifiable from anything codex guarantees. I've tried to avoid making the same kind of bet — see below.

## The shape I have working

Claude's `stream-json` already distinguishes these sources, so this can be read rather than inferred. `parseStream` currently merges assistant text, result metadata and non-JSON stderr into one `RecentText` blob. Splitting it into two windows with different authority:

- **`DiagnosticText`** — structured records only: `result` with `is_error:true`, assistant messages with `is_api_error_message` or a non-empty `error`, relevant `system` events, plus non-JSON CLI lines. Always pattern authority.
- **`RecentText`** — the surfaced narration. Kept for display, and promoted to pattern authority only when the process exited non-zero or hit the idle timeout **without** emitting a structured ralphex signal.

That last clause is the one heuristic left, and I'd rather flag it than have you find it: it exists so wrappers that surface real failures as ordinary deltas keep working. If you'd rather diagnostics be the only authority ever, that's a smaller change and I'm happy to do it that way — it just risks those wrappers.

One wrinkle worth knowing about: `system/api_retry` records have to be provisional. They stay authoritative if the stream ends without a terminal result, but a later `result` with `is_error:false` retires them — otherwise a retry claude recovered from internally would rerun completed work.

While capturing fixtures against Claude Code 2.1.251 I also hit something that seems worth recording either way: **a rate-limit or auth failure can arrive with `subtype:"success"` and `is_error:true`**. So `is_error` and the error fields are authoritative and `subtype` is not. Inferring success from `subtype` alone would be wrong today.

## What I'm asking

I have this implemented and tested (276 lines of production code in `executor.go`, plus fixtures minimized from real 2.1.251 captures and a regression matrix covering narrated-vs-genuine across clean and non-zero exits). Codex and custom executors are untouched.

But I'd rather not open a PR against code you've already defended without checking the shape first. Would you like:

1. the two-window split as described,
2. diagnostics-only authority with no failed-run fallback (smaller, slightly riskier for wrappers),
3. or something else entirely?

Happy to write it whichever way you prefer, or to drop it if you think the config route really does cover this and I've missed how.

---

Environment: ralphex 0.20.0 (built from master), Windows 11, claude executor, `wait_on_limit` set, Claude Code 2.1.251.

yeroo:
Fixes #452. Option 1, the two-source split, with the four points from your reply folded in.

**194 lines of production code, all in `pkg/executor/executor.go`.** The rest is 627 lines of tests, 7 fixtures and docs.

## The change

`parseStream` merged assistant text, result metadata and non-JSON stderr into one `RecentText` blob, and `Run` matched patterns against it even on a clean zero exit. Split into two windows with different authority:

- **`DiagnosticText`** — result records that authenticate themselves as errors, assistant API-error messages, system error events, plus non-JSON CLI lines. Always pattern authority.
- **`RecentText`** — surfaced narration, kept for display. Promoted to pattern authority only when the process itself failed.

`streamEvent` gained the fields that carry the provenance and that ralphex previously parsed none of: `is_error`, `error`, `error_status`, `api_error_status`, `is_api_error_message`, `terminal_reason`.

## Your four points

**Promotion rule is exit status + stream error, not signal absence.** You were right that `Signal == ""` was unusable — `review_first.txt:84-86` path B stays silent by design and `task.txt:43` only emits `ALL_TASKS_DONE` once the plan is empty. It now reads:

```go
patternText := result.DiagnosticText
if outcome.waitErr != nil || result.Error != nil {
    patternText += "\n" + result.RecentText
}
```

An idle timeout cancels the stream context and `readLines` returns a wrapped `ctx.Err()`, so it reaches promotion through `result.Error` rather than needing its own clause. Two matrix cases that asserted a signal protected narration are inverted accordingly, and `TestClaudeExecutor_Run_WaitError_WithSignalAndNarratedErrorPattern` became `..._WaitErrorPromotesNarrationEvenWithSignal`.

**Result-text capture is gated on error authentication.** `isErrorResult` is `is_error:true || terminal_reason == "api_error" || api_error_status != 0`. A bare `terminal_reason` does not qualify — there is a regression test for a successful summary carrying `terminal_reason:"stop_sequence"` and quoting a configured phrase, which must stay out. `subtype` is never consulted.

**`system/api_retry` is out entirely**, along with the provisional/retire machinery. No 2.1.251 fixture shows it as sole carrier — in `diagnostic-authentication.jsonl` the assistant and result records carry the same 401 — so it was buying nothing and `textWindow` is a plain ring again. This is why the production diff came in smaller than the 276 lines I quoted in the issue.

**Codex and custom untouched**, with assertions in both suites that `DiagnosticText` stays empty on their paths. Fixtures are included, minimized from real Claude Code 2.1.251 captures including the `subtype:"success"` + `is_error:true` rate limit and the 401 authentication case.

## Testing

`go test ./pkg/executor/... ./pkg/processor/...` passes.

One caveat worth stating plainly: I develop on Windows, and on a **clean master** `TestCodexExecutor_findRolloutFile` and `TestCodexExecutor_tailRolloutFile_streamsAssistantMessages` already fail there, because they set `HOME` while `os.UserHomeDir` reads `USERPROFILE` on Windows. That is unrelated to this branch and I deliberately kept the fix out of it — happy to send it separately if useful. Everything else in both packages is green, and CI will cover the Linux path.

I could not reproduce a real session limit to capture end-to-end, which is why the fixtures matter.

https://claude.ai/code/session_01RDQ6n52k9GvTpyqUHMQf5W


umputun (approved):
lgtm

PR #459, opened 2026-09-14
Fix watcher tests leaking tailers on Windows

LexTalionis88:
## Summary
- close test SessionManager instances so watcher-created tailers release progress files
- close the directly-created session in the stored-offset watcher regression test

## Tests
- go test ./pkg/web
- make test (pkg/web passes; existing Windows/env failures remain in other packages)


umputun (approved):
lgtm

PR #460, opened 2026-09-15
Make tests portable on Windows

LexTalionis88:
## Summary
- replace Unix-only test commands and paths with portable helpers
- skip chmod-based permission tests on Windows and skip symlink tests when privileges are unavailable
- make progress history stems and coverage post-processing portable on Windows

## Tests
- make test


umputun:
closing this one. Linux and macOS are the supported platforms, and the README sets the bar for Windows patches: clear-cut cause, small, self-contained, and no effect on Linux or macOS. This PR is 17 files, it changes production code (the history stem rewrite in `pkg/progress`, which also folds distinct plans into one archive directory for names with a leading dot or a colon) and the Makefile, and the new editor fixture fails on macOS when `TMPDIR` contains a space. That is a lot of surface for a Windows workflow I can't run or verify here, since there is no Windows CI and I have no Windows machine.

